PR #8, opened 2022-12-15
Update resource creation logic with patch

ianstanton:
Use `patch` instead of `create` when creating stateful set. By using `patch`, we no longer need to check for an existing stateful set.


ChuckHend (approved):
Awesome!

PR #10, opened 2022-12-16
Examples

ryw:


examples/instances/read.toml
```diff
@@ -0,0 +1,10 @@
+[instance]
+name = "primary"
```
ianstanton, line 2:
```suggestion
name = "read"
```

ianstanton (commented):
This name should be updated. @ryw go ahead and commit this suggestion if you agree.

ChuckHend (approved):

PR #12, opened 2022-12-16
Pg metrics

ChuckHend:
Sets pattern for exporting pg server metrics. two bugs currently exist:
1. background thread doesn't get killed when pgx stops
2. the query result seems to have some strange type casting. e.g. expecting a value ~100 and getting some very large int like 4974814424

### DIFF
--- a/extensions/prometheus_exporter/src/webserver.rs
+++ b/extensions/prometheus_exporter/src/webserver.rs
@@ -1,12 +1,23 @@
 // https://github.com/prometheus/client_rust/blob/master/examples/actix-web.rs
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
+use std::{thread, time};
 
 use actix_web::{web, App, HttpResponse, HttpServer, Responder, Result};
+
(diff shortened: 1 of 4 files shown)


extensions/prometheus_exporter/src/webserver.rs
```diff
@@ -41,31 +69,85 @@ pub async fn metrics_handler(state: web::Data<Mutex<AppState>>) -> Result<HttpRe
...
+    let uptime = Arc::new(Mutex::new(i64::default()));
+    let clone = Arc::clone(&uptime);
+    println!("query called");
+    // BackgroundWorker::transaction(move || {
+    //     let ut: Option<i64> = Spi::get_one(UPTIME_QUERY);
+    //     match ut {
+    //         Some(t) => println!("t: {:?}", t),
+    //         None => println!("no value"),
+    //     }
+    // });
```
ianstanton, line 101:
```suggestion
```

ianstanton (approved):

PR #15, opened 2022-12-17
Add pgexporter tests

ChuckHend:
- Adds a test for the exporter. Very basic for now and only testing on pg14 (until we can figure out a better way to setup the test tooling).

- Adds a github workflow to run the above test

### DIFF
--- a/.github/workflows/postgres_extension.yml
+++ b/.github/workflows/postgres_extension.yml
@@ -3,6 +3,9 @@ name: Prometheus Exporter for Postgres
 on:
   pull_request:
     branches: [ main ]
+    paths: 
+    - '.github/workflows/postgres_extension.yml'
+    - 'extensions/**'
 
 jobs:
   build_and_test:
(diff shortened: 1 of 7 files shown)


.github/workflows/postgres_extension.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: setup pgx
+        run: cargo install --locked cargo-pgx
+
+      - name: init
+        run: cargo pgx init
```
ChuckHend, line 25: this takes a lot of time because it downloads and compiles several versions of postgres. ideally we could prebuild these, maybe find or build a utility image that has several versions built.

ChuckHend: on second thought, its only slow compared to what I expected.  The `pgx` project's CI is a ~ 25 minute job, so maybe it isn't going to be a problem: https://github.com/tcdi/pgx/actions

ianstanton (approved):

PR #20, opened 2022-12-19
[COR-37] Add unit testing with mocked Kubernetes API

sjmiller609:


coredb-operator/src/fixtures.rs
```diff
@@ -0,0 +1,150 @@
...
+pub struct ApiServerVerifier(Handle<Request<Body>, Response<Body>>);
+
+/// Create a responder + verifier object that deals with the main reconcile scenarios
+///
+/// 1. objects without finalizers will get a finalizer applied (and not call the apply loop)
+/// 2. finalized objects will run through the apply loop
+/// 3. finalized objects "with errors" (i.e. the "illegal" object) will short circuit the apply loop
+/// 4. objects with a deletion timestamp will run the cleanup loop (which will send an event)
+impl ApiServerVerifier {
+    pub fn handle_finalizer_creation(self, coredb_: &CoreDB) -> JoinHandle<()> {
```
sjmiller609, line 45: These tests act like coroutines to the code in the controller. They mock the k8s API, and on each call to "next_request" it will wait on the request to come from the controller code. Then it handles the mocked k8s logic in these functions, then return a result to the controller code.

coredb-operator/src/fixtures.rs
```diff
@@ -0,0 +1,150 @@
...
+        let coredb = coredb_.clone();
+        tokio::spawn(async move {
+            pin_mut!(handle);
+            let (request, send) = handle.next_request().await.expect("service not called");
+            // We expect a json patch to the specified coredb adding our finalizer
+            assert_eq!(request.method(), http::Method::PATCH);
+            assert_eq!(
+                request.uri().to_string(),
+                format!("/apis/kube.rs/v1/namespaces/testns/coredbs/{}?", coredb.name_any())
+            );
```
sjmiller609, line 56: For example here, the first request that's received is the PATCH to the CoreDB resource, which occurs in the controller here:

https://github.com/CoreDB-io/coredb/blob/7e3237fa9c9698a547f6c7efa7cc29ddb8748f4d/coredb-operator/src/controller.rs#L93-L104

coredb-operator/src/fixtures.rs
```diff
@@ -0,0 +1,150 @@
...
+            let (request, send) = handle.next_request().await.expect("service not called");
+            assert_eq!(request.method(), http::Method::POST);
+            assert_eq!(
+                request.uri().to_string(),
+                format!("/apis/events.k8s.io/v1/namespaces/testns/events?")
+            );
+            // pass the event straight through
+            send.send_response(Response::builder().body(request.into_body()).unwrap());
+        })
+    }
```
sjmiller609, line 84: I am reviewing this part, I think that in the example controller, they are publishing events, but not in our case. So I think I should delete this part.

https://github.com/kube-rs/controller-rs/blob/e20529654060d8bbd067be59a828f6f6f7881cba/src/controller.rs#L94-L106

coredb-operator/src/fixtures.rs
```diff
@@ -0,0 +1,150 @@
...
+            // After the PATCH to CoreDB, we expect a PATCH to StatefulSet
+            let (request, send) = handle
+                .next_request()
+                .await
+                .expect("Kube API called to PATCH StatefulSet");
+            assert_eq!(request.method(), http::Method::PATCH);
+            assert_eq!(
+                request.uri().to_string(),
+                format!("/apis/apps/v1/namespaces/testns/statefulsets/testdb?&force=true&fieldManager=cntrlr")
+            );
```
sjmiller609, line 127: In this test, we first get the PATCH to the CoreDB object, then wait for the next request, and check that it's to PATCH the corresponding StatefulSet.

coredb-operator/src/main.rs
```diff
@@ -44,8 +44,13 @@ async fn main() -> Result<()> {
     // Initialize tracing
     tracing::subscriber::set_global_default(collector).unwrap();
 
+    // Initialize the Kubernetes client
+    let client_future = kube::Client::try_default();
+    let client = match client_future.await {
+        Ok(wrapped_client) => wrapped_client,
+        Err(error) => panic!("Please configure your Kubernetes Context"),
+    };
```
sjmiller609, line 52: I added this change to print an informative error message when kubernetes context is not configured.

coredb-operator/src/fixtures.rs
```diff
@@ -0,0 +1,150 @@
...
+use tokio::task::JoinHandle;
+use tower_test::mock::{self, Handle};
+
+impl CoreDB {
+
+    /// A normal test CoreDB
+    pub fn test() -> Self {
+        let mut d = CoreDB::new("testdb", CoreDBSpec::default());
+        d.meta_mut().namespace = Some("testns".into());
+        d.meta_mut().uid = Some("752d59ef-2671-4890-9feb-0097459b18c8".into());
```
sjmiller609, line 19: this line was different than in the example code because we are using ownership reference from CoreDB to StatefulSet, we need the CoreDB to have a uid so this function works: https://docs.rs/kube-core/0.77.0/src/kube_core/resource.rs.html#105

In our create_sts function while it's handling the mocked response.

.github/workflows/test_operator.yml
```diff
@@ -0,0 +1,36 @@
...
+  lint:
+    name: Linting
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+      - name: Cargo format check
+        run: cargo +nightly fmt --check
```
ianstanton: Thanks for adding this!

sjmiller609:
just getting the lint step to work in ci

ianstanton (approved):
Great work! Will we be able to do things like:
- Assert some resource is created successfully (looks like we're checking status for that here)
- Assert expected spec values are present on some resource
- Assert something happens to a resource when CoreDB spec values change

Similar to the examples given [here](https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html?highlight=ginkgo#testing-your-controllers-behavior) (golang test framework used in other operators).

PR #21, opened 2022-12-20
Update Controller RBAC and README

ianstanton:
- Allow for controller deployment to manage statefulsets
- Update README with more accurate local dev instructions. In the future we should add some steps to our `justfile` for creating a kind cluster + local registry. This will simplify local development for us.

### DIFF
--- a/coredb-operator/README.md
+++ b/coredb-operator/README.md
@@ -16,7 +16,7 @@ As an example; install [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start/#
 Apply the CRD from [cached file](yaml/crd.yaml), or pipe it from `crdgen` (best if changing it):
 
 ```sh
-cargo +nightly run --bin crdgen | kubectl apply -f -
+cargo run --bin crdgen | kubectl apply -f -
 ```
 
 ### Opentelemetry (optional)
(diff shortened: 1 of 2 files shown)


sjmiller609 (approved):
which dependency is requiring nightly build?

ianstanton:
> which dependency is requiring nightly build?

@sjmiller609 This is actually no longer necessary (it was last week, not sure what changed to fix this). I'll update these steps.

sjmiller609:
> > which dependency is requiring nightly build?
> 
> @sjmiller609 This is actually no longer necessary (it was last week, not sure what changed to fix this). I'll update these steps.

Ok cool, something I just noticed in my PR is that the nightly build does more linting auto-corrections, and I think it works ok to use the nightly features on this lint step even though using the stable compiler.

```
➜  coredb-operator git:(COR-37) rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
➜  coredb-operator git:(COR-37) cargo +nightly fmt
➜  coredb-operator git:(COR-37) $?
zsh: command not found: 0
```

https://github.com/CoreDB-io/coredb/pull/20/commits/c1f45bedb5246eb748bf7b967fde71802f76dd63

PR #23, opened 2022-12-20
[COR-37] Integration / functional tests

sjmiller609:
- add integration testing setup
  - starts Kind cluster(s)
  - installs CRD
  - starts operator
  - runs integration tests
- integration tests:
  - apply a configuration of coreDB
  - check pod becomes running

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -90,6 +90,7 @@ jobs:
           set -xe
           # Install the CRD
           cargo run --bin crdgen | kubectl apply -f -
+          kubectl get crds
           # Start the operator in the background
           cargo run &
           # Run the tests
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -64,14 +64,15 @@ jobs:
       matrix:
         # Go here for a list of versions:
         # https://github.com/kubernetes-sigs/kind/releases
-        k8s_version:
-          - 'v1.25.3'
+        node_image:
+          - 'kindest/node:v1.25.3'
+          - 'kindest/node:v1.22.15'
     steps:
       - uses: actions/checkout@v2
       - name: Create k8s Kind Cluster
-        uses: helm/kind-action@v1.4.0
+        uses: helm/kind-action@v1.5.0
         with:
-          version: ${{ matrix.version }}
+          node_image: ${{ matrix.node_image }}
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -82,4 +82,4 @@ jobs:
           workspaces: |
             coredb-operator
       - name: Run functional / integration tests
-        run: cargo test -- --ignored
+        run: cargo test -- --ignored --nocapture
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -19,25 +19,6 @@ on:
       - release**
 
 jobs:
-  test:
-    name: Run unit tests
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-      - uses: Swatinem/rust-cache@v2
-        with:
-          prefix-key: "coredb-operator-test"
-          workspaces: |
-            coredb-operator
-      - name: Unit tests
-        run: cargo test
-      # WIP: If the unit tests pass, also run integration tests
-      # - name: Integration tests
-      #   run: cargo test -- --ignored
-
   lint:
     name: Run linters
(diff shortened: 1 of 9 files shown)


coredb-operator/tests/integration_tests.rs
```diff
@@ -0,0 +1,68 @@
...
+            "name": "sample-coredb"
+        },
+        "spec": {
+            "replicas": 1
+        }
+    });
+    let params = PatchParams::apply("coredb-integration-test");
+    let patch = Patch::Apply(&coredb_json);
+    let coredb_resource = coredbs.patch("sample-coredb", &params, &patch).await;
+    panic!("{:?}", coredb_resource)
```
sjmiller609, line 36: got it working to create and panic here for debugging the resource. now looking into waiting - since need some logic like "check that a pod starts and becomes ready in the next 60 seconds" or something

sjmiller609: I was originally going down a too-deep path, I think this is the most abstract API available https://docs.rs/kube/latest/kube/runtime/wait/index.html

sjmiller609: I think I got that part working

coredb-operator/tests/integration_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+    use kube::{
+        api::{Patch, PatchParams},
+        runtime::wait::{await_condition, conditions},
+        Api, Client, Config,
+    };
+    use rand::Rng;
+
+    #[tokio::test]
+    #[ignore]
+    async fn functional_test_basic_create() {
```
sjmiller609, line 26: this test deploys a CoreDB into the cluster we are connected to, then checks that a pod starts

.github/workflows/operator.yml
```diff
@@ -0,0 +1,113 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      # fail-fast means to cancel all jobs if one fails
+      fail-fast: false
+      matrix:
+        # Go here for a list of versions:
+        # https://github.com/kubernetes-sigs/kind/releases
+        node_image:
+          - 'kindest/node:v1.25.3'
+          - 'kindest/node:v1.22.15'
```
sjmiller609, line 69: we can test against multiple k8s versions

coredb-operator/tests/integration_tests.rs
```diff
@@ -0,0 +1,119 @@
...
+            std::time::Duration::from_secs(timeout_seconds),
+            await_condition(pods, &pod_name, conditions::is_pod_running()),
+        )
+        .await
+        {
+            panic!(
+                "\n\nERROR: Did not find the pod {} to be running after waiting for {} seconds\n\n",
+                pod_name, timeout_seconds
+            )
+        }
```
sjmiller609, line 77: I found this pattern to wait on something to be useful

coredb-operator/tests/integration_tests.rs
```diff
@@ -0,0 +1,119 @@
...
+        // Check that the CRD is installed
+        let crds: Api<CustomResourceDefinition> = Api::all(client.clone());
+        if let Err(_) = tokio::time::timeout(
+            std::time::Duration::from_secs(2),
+            await_condition(crds, "coredbs.kube.rs", conditions::is_crd_established()),
+        )
+        .await
+        {
+            panic!("\n\nERROR: Did not find the CRD to be installed.\n\n")
+        }
```
sjmiller609, line 115: For example same pattern here to wait on CRD. kube-rs also offers custom conditions, so can write a function and wait until that is satisfied.

https://docs.rs/kube/latest/kube/runtime/wait/trait.Condition.html

ChuckHend (approved):

coredb-operator/tests/integration_tests.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    async fn kube_client() -> kube::Client {
+        // Initialize the Kubernetes client
+        let client_future = Client::try_default();
+        let client = match client_future.await {
+            Ok(wrapped_client) => wrapped_client,
+            Err(_error) => panic!("Please configure your Kubernetes Context"),
+        };
+        // Get the name of the currently selected namespace
+        let selected_namespace = Config::infer().await.unwrap().default_namespace;
```
clux, line 88: NB: you are technically doing `Config::infer` twice in this block. you could use the result of the config inference to create a `Client`

coredb-operator/tests/integration_tests.rs
clux: if you are just panic'ing you could do `tokio::timeout(...).await.expect("CRD was not installed in time")` for a "one-liner" - also passes error info into the panic

coredb-operator/tests/integration_tests.rs
sjmiller609: I will check on this. Thank you!

PR #25, opened 2022-12-21
[COR-49] Add linting and dependency caching for prometheus exporter

sjmiller609:
While it's top of mind from the other change, I figure to include some things I learned from the operator tests here.


coredb-operator/tests/integration_tests.rs
```diff
@@ -40,7 +40,7 @@ mod test {
         let replicas = 1;
 
         // Timeout settings while waiting for an event
-        let timeout_seconds = 30;
+        let timeout_seconds = 60;
```
sjmiller609, line 43: this change is not related to COR-49, it's just to address this comment: https://github.com/CoreDB-io/coredb/pull/24#issuecomment-1362081997

sjmiller609: also, this is why the operator tests are running in this PR.

.github/actions/pgx-init/action.yml
```diff
@@ -0,0 +1,41 @@
+name: 'pgx initialization'
```
sjmiller609, line 1: This action will perform the pgx initialization steps

.github/actions/pgx-init/action.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          set -x
+          ls -ltrah /home/runner/.pgx || true
+          # pgx init can take a long time, and it re-compiles postgres even when there
+          # is a cached version. So, we can just check for the directory and
+          if find /home/runner/.pgx | grep $(awk -F "=" '/pg$${{ inputs.pg_version }}/ {print $2}' /home/runner/.pgx/config.toml | tr -d '"'); then
+            echo "Already found pgx is initialized. Skipping 'cargo pgx init' command."
+          else
+            cargo pgx init --pg${{ inputs.pg_version }} download
+          fi
```
sjmiller609, line 31: The logic here is to avoid running pgx init on a cache hit. The reasoning of the logic is here in the pgx discord https://discordapp.com/channels/561648697805504526/749454413407584258/1055239589276700795

.github/actions/pgx-init/action.yml
```diff
@@ -0,0 +1,32 @@
...
+          set -x
+          ls -ltrah /home/runner/.pgx || true
+          # pgx init can take a long time, and it re-compiles postgres even when there
+          # is a cached version. So, we can just check for the directory and
+          if find /home/runner/.pgx | grep $(awk -F "=" '/pg${{ inputs.pg_version }}/ {print $2}' /home/runner/.pgx/config.toml | tr -d '"'); then
+            echo "Already found pgx is initialized. Skipping 'cargo pgx init' command."
+          else
+            cargo pgx init --pg${{ inputs.pg_version }} download
+          fi
+          ls -ltrah /home/runner/.pgx
```
ChuckHend, line 32: Is this checking to see if the node running our github workflow is the same node already had `cargo pgx init` ran on it? I figured we'd eventually have pre-built images and run these tests on them. That way we'd never have to run `pgx init` if we dont want to.

sjmiller609: In a previous step it does a cache restore attempt, so it's checking for cache hit or miss.

Currently understanding if there is something I'm missing, for example postgres started by init that I didn't notice, i.e. if something is needed besides files that can be cached.

Also, for some reason when the tests run, I am having it fail sometimes in CI, and also slow. But on my laptop it's super fast and never fails, but I'm giving more tries. Looks like this when it fails. https://github.com/CoreDB-io/coredb/actions/runs/3753427332/jobs/6376678527

sjmiller609: I currently suspect it's related to resources and there is some subtle race condition when limited on resources, but not sure yet.

sjmiller609: the location of PGDATA is

`~/.pgx/data-14/`

that includes `postgresql.auto.conf`

.github/workflows/prom_exporter_ext.yml
```diff
@@ -11,30 +11,81 @@ on:
...
+          set -xe
+          # Debugging information
+          ls -ltrah /home/runner/.pgx/
+          ls -ltrah /home/runner/.pgx/data-${{ matrix.pg_version }}/
+          whoami
+          cat /home/runner/.pgx/data-${{ matrix.pg_version }}/postgresql.conf
+          cat /home/runner/.pgx/data-${{ matrix.pg_version }}/postgresql.auto.conf
+          ls ./target/target/pgx-test-data-${{ matrix.pg_version }}
+          # ensure test data directory is deleted before starting tests
+          rm -rf ./target/target/pgx-test-data-${{ matrix.pg_version }}
```
sjmiller609, line 87: I'm having some issue with failure in pg_test to modify the config file. I found log from working case that the data directory is here:
```
target/pgx-test-data-14
```

so, we can try to delete it before running the test, this might help. The working logs include creating the directory:

```
The files belonging to this database system will be owned by user "runner".
This user must also own the server process.

The database cluster will be initialized with locale "C.UTF-8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory /home/runner/work/coredb/coredb/extensions/prometheus_exporter/target/pgx-test-data-14 ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... Etc/UTC
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
initdb: warning: enabling "trust" authentication for local connections
syncing data to disk ... ok

You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.

Success. You can now start the database server using:

    /home/runner/.pgx/14.6/pgx-install/bin/pg_ctl -D /home/runner/work/coredb/coredb/extensions/prometheus_exporter/target/pgx-test-data-14 -l logfile start

test webserver::tests::pg_test_pg_uptime ... ok
```

sjmiller609: ok I think that worked

.github/actions/pgx-init/action.yml
sjmiller609: Hey I was off topic in this conversation, I had broken something with the cache restore of the target/ directory. Yes I do think docker image seems like a good way to go about it.

ChuckHend: Thanks for doing this. These tests were super slow. And even slower at first when we were running `init` on all pg 11-15.

ChuckHend (approved):

PR #27, opened 2022-12-21
[COR-50] Refactor Prometheus Exporter

ChuckHend:
- Move metric structs to a new module
- Add utility function for executing SQL, which will get reused for a lot of the metrics which have yet to be implemented
- All code concerned with the SPI will leave in query.rs


extensions/prometheus_exporter/src/metrics.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    let metrics = metrics_clone.lock().unwrap();
+    while !term.load(Ordering::Relaxed) {
+        let uptime: i64 = query::handle_query(query::UPTIME_QUERY).unwrap();
+        {
+            metrics.pg_uptime(uptime);
+            thread::sleep(time::Duration::from_millis(2500));
+        }
+    }
+}
```
ChuckHend, line 40: I think there's still some work that can be done here. Maybe every metric value could have a trait associated with that contains the logic needed to update its value.

extensions/prometheus_exporter/src/metrics/query.rs
```diff
@@ -0,0 +1,38 @@
...
+        let mut obj_clone = clone.lock().unwrap();
+        *obj_clone = query_exec(&query).unwrap();
+    });
+    let x = Some(*uptime.lock().unwrap());
+    x
+}
+
+fn query_exec(query: &str) -> Option<i64> {
+    Spi::get_one(&query)
+}
```
ChuckHend, line 26: trying to break up `handle_query` more so it can be tested. Still trying to figure out how to configure `pgx` to run tests on `BackgroundWorker` though...thats up next.

sjmiller609 (approved):

PR #32, opened 2023-01-01
Reformat futures import

EvanHStanton:
Hello, I believe this import involves two sub-modules from the futures crate, i.e., future and stream. Expanded, I understand it would look like:

```
futures::future::BoxFuture
futures::future::FutureExt
futures::stream::StreamExt
```

Very enthusiastic about your work.

-Evan


ianstanton (approved):
Looks good to me! This looks to be doing the same thing, just in a more explicit manner. Thanks for the PR.

ChuckHend (approved):

PR #34, opened 2023-01-03
Updated versions of dependencies

EvanHStanton:
- Updated dependencies.

@ChuckHend: I was interested in this particular Cargo.toml file due to its use of the object relational mapper crate, ```diesel```. If you have not already come across the ```sea-orm``` crate, it might be interesting to take a quick peek. It is an alternative that is built from the ground up with async in mind, and built upon ```sqlx```, which the ```sea-orm``` docs notes works well with ```actix```, ```async-std```, and ```tokio```.

https://www.sea-ql.org/SeaORM/

```diesel``` might also check these boxes, and it should be noted that there is an extension called ```diesel-async```, which may address well one of the critiques, i.e., that it wasn't originally built with async in mind.

### DIFF
--- a/extensions/dev/http_webserver/benchmarking/Cargo.toml
+++ b/extensions/dev/http_webserver/benchmarking/Cargo.toml
@@ -5,16 +5,16 @@ edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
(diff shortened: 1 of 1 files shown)


ChuckHend (approved):
Awesome! Thanks for pointer to sea-orm and updating the deps!

ianstanton (approved):
:boom:

PR #40, opened 2023-01-10
[COR-102] Run postgres pod as 'postgres' user

sjmiller609:
Since the container we are using, for now, was intended to run in Docker in a single "run" command, it starts as root, adjusts the permissions on the directory, then switches into postgres user to start postgres.

When working on getting psql commands to run using 'kubectl exec', I noticed an issue where by default it will exec with the root user because of this initialization step. In this PR, I refactored the initialization into an initContainer.

Also, I adjusted the test to assert the container is Ready. I noticed that the test previously only checked the container is running, but test would pass if the Pod then crashed.

### DIFF
--- a/coredb-operator/src/statefulset.rs
+++ b/coredb-operator/src/statefulset.rs
@@ -41,8 +41,7 @@ fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
     let postgres_container = Container {
         env: postgres_env.clone(),
         security_context: Some(SecurityContext {
-            //run_as_user: Some(cdb.spec.uid.clone() as i64),
-            run_as_user: Some(999),
+            run_as_user: Some(cdb.spec.uid.clone() as i64),
             allow_privilege_escalation: Some(false),
             ..SecurityContext::default()
         }),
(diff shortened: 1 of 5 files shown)


coredb-operator/src/statefulset.rs
```diff
@@ -26,6 +27,61 @@ pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error>
...
+    }]);
+
+    // This container for running postgresql
+    let postgres_container = Container {
+        env: postgres_env.clone(),
+        security_context: Some(SecurityContext {
+            run_as_user: Some(999),
+            allow_privilege_escalation: Some(false),
+            ..SecurityContext::default()
+        }),
```
sjmiller609, line 49: In the docker container we are using, postgres user is ID 999.

coredb-operator/src/statefulset.rs
```diff
@@ -26,6 +27,61 @@ pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error>
...
+        name: "data".to_owned(),
+        mount_path: "/var/lib/postgresql/data".to_owned(),
+        ..VolumeMount::default()
+    }]);
+
+    // This container for running postgresql
+    let postgres_container = Container {
+        env: postgres_env.clone(),
+        security_context: Some(SecurityContext {
+            run_as_user: Some(999),
```
ianstanton, line 46: Would we ever want to change this? If yes, we should add a spec value like `UID`.

sjmiller609: Good call, added that plus a test for it

coredb-operator/src/statefulset.rs
```diff
@@ -26,6 +27,61 @@ pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error>
...
+    set -e
+    source /usr/local/bin/docker-entrypoint.sh
+    set -x
+    docker_setup_env
+    docker_create_db_directories
+                        "
+            .to_string(),
+        ]),
+        ..Container::default()
+    };
```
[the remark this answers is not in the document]
sjmiller609: Ok I think I covered this, but I want to check in on it with you later

PR #41, opened 2023-01-12
[COR-102] Add ability for operator to run psql commands

sjmiller609:
Add a function to CoreDB implementation that allows for calling psql commands in the primary postgres container. The kube API exec feature is used to connect to the container.


coredb-operator/Cargo.toml
```diff
@@ -27,7 +27,7 @@ telemetry = ["tonic", "opentelemetry-otlp"]
 [dependencies]
 actix-web = "4.2.1"
 futures = "0.3.25"
-tokio = { version = "1.22.0", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.22.0", features = ["macros", "rt-multi-thread", "io-util"] }
```
sjmiller609, line 30: read_to_string Tokio

https://docs.rs/tokio/0.2.9/tokio/fs/fn.read_to_string.html

sjmiller609:
debugging a flake, found this error message.

```
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
Is the server running locally and accepting connections on that socket?
```

that's the message you get when pg not running. something like it's restarting the DB right after it starts? especially since the command before that exits in success oddly

sjmiller609:
it's just because pg restarts when the container starts because that's how the entrypoint of the PG container we are using is written. will adjust readiness of pod to account for that

coredb-operator/src/statefulset.rs
```diff
@@ -96,8 +54,55 @@ fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
             },
             template: PodTemplateSpec {
                 spec: Some(PodSpec {
-                    containers: vec![postgres_container],
-                    init_containers: Option::from(vec![postgres_init_container]),
+                    containers: vec![
```
sjmiller609, line 57: moved container definitions back into statefulset https://github.com/CoreDB-io/coredb/pull/40#discussion_r1066116171

sjmiller609: stuff shared between the container and the init container is still outside to avoid duplicating it (e.g. volume mount)

sjmiller609:
running it a few times to check there are not flakes

coredb-operator/src/controller.rs
```diff
@@ -139,6 +147,50 @@ impl CoreDB {
...
+
+    async fn primary_pod(&self, client: Client) -> Result<Pod> {
+        let sts = stateful_set_from_cdb(self);
+        let sts_name = sts.metadata.name.unwrap();
+        let sts_namespace = sts.metadata.namespace.unwrap();
+        let label_selector = format!("statefulset={}", sts_name);
+        let list_params = ListParams::default().labels(&label_selector);
+        let pods: Api<Pod> = Api::namespaced(client, &sts_namespace);
+        let pods = pods.list(&list_params);
+        // For the time being, we assume that the first pod is the primary
```
sjmiller609, line 159: At the moment, we are just running one replica to avoid needing to handle service routing to the primary. We will need some kind of logic to pick the primary pod and point the service at it once we are doing standby / failover pods.

coredb-operator/tests/integration_tests.rs
```diff
@@ -56,7 +57,7 @@ mod test {
 
         // Timeout settings while waiting for an event
         let timeout_seconds_start_pod = 60;
-        let timeout_seconds_pod_ready = 10;
+        let timeout_seconds_pod_ready = 30;
```
sjmiller609, line 60: now that there is a readiness probe, it doesn't just mark the pod as ready instantly when it's started

sjmiller609: On my local, the pod is 15s old when it becomes ready.

coredb-operator/src/controller.rs
```diff
@@ -139,6 +147,50 @@ impl CoreDB {
...
+        let pods: Api<Pod> = Api::namespaced(client.clone(), &self.metadata.namespace.clone().unwrap());
+        let mut attached_process = pods.exec(&pod_name, psql_command, &attach_params).await.unwrap();
+        // attached_process.join();
+        let mut stdout_reader = attached_process.stdout().unwrap();
+        let mut result_stdout = String::new();
+        stdout_reader.read_to_string(&mut result_stdout).await.unwrap();
+        let mut stderr_reader = attached_process.stderr().unwrap();
+        let mut result_stderr = String::new();
+        stderr_reader.read_to_string(&mut result_stderr).await.unwrap();
+        return Ok(result_stdout + &result_stderr);
```
sjmiller609, line 192: I think the return type can be further improved, but I need to learn more rust. would be better if the result is like

```
result.stdout
result.stderr
result.exit_code
```

sjmiller609: at the moment it's just concatenating stdout and stderr

coredb-operator/tests/integration_tests.rs
```diff
@@ -99,6 +100,37 @@ mod test {
             "Did not find the pod {} to be ready after waiting {} seconds",
             pod_name, timeout_seconds_pod_ready
         ));
+        println!("Found pod ready: {}", pod_name);
```
sjmiller609, line 103: I am just putting this additional test logic into the existing test to avoid launching two coreDBs per test run, but I think there is room for improvement by setting up test fixture to allow for one, or limited set, of CoreDB to be shared between tests that would be compatible running on a shared target.

sjmiller609: This test addition is just checking we can run psql commands and the output looks like expected on a few basic commands.

coredb-operator/Cargo.toml
```diff
@@ -51,5 +51,5 @@ rand = "0.8.5"
 tower-test = "0.4.0"
 
 [dependencies.kube]
-features = ["runtime", "client", "derive"]
+features = ["runtime", "client", "derive", "ws"]
```
sjmiller609, line 54: need for exec feature, I assume this means websockets

coredb-operator/src/statefulset.rs
```diff
@@ -127,6 +132,21 @@ fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
     return sts;
 }
 
+pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error> {
```
sjmiller609, line 135: just moved this block so the tests are at the bottom of the file

coredb-operator/src/statefulset.rs
```diff
@@ -96,8 +54,55 @@ fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
...
+                                ..ContainerPort::default()
+                            }]),
+                            volume_mounts: postgres_volume_mounts.clone(),
+                            readiness_probe: Some(Probe {
+                                exec: Some(ExecAction {
+                                    command: Some(vec![String::from("pg_isready")]),
+                                }),
+                                initial_delay_seconds: Some(3),
+                                ..Probe::default()
+                            }),
```
sjmiller609, line 79: this is new logic to add a readiness probe, using the postgres feature "pg_isready"

sjmiller609:
I'm also having Darren of cincinnati ventures review it

sjmiller609:
some notes from peering with Darren:

Mods in the lib.rs should go at top of file

Implement CoreDB should be directly under the CoreDBSpec

RENAME kube.rs to coredb.io


What about test fixtures?

Lazy static inside test module?

Rstest?

coredb-operator/tests/integration_tests.rs
sjmiller609: I think this can be accomplished with either lazy static or with rstest

coredb-operator/src/controller.rs
```diff
@@ -139,6 +147,50 @@ impl CoreDB {
             .map_err(Error::KubeError)?;
         Ok(Action::await_change())
     }
+
+    async fn primary_pod(&self, client: Client) -> Result<Pod> {
```
sjmiller609, line 151: When I'm returning Result, I ought to also handle errors. And when I'm unwrapping, I should pass errors up with '?'

coredb-operator/src/controller.rs
```diff
@@ -139,6 +147,50 @@ impl CoreDB {
...
+        let pod_name = self
+            .primary_pod(client.clone())
+            .await
+            .unwrap()
+            .metadata
+            .name
+            .unwrap();
+        let pods: Api<Pod> = Api::namespaced(client.clone(), &self.metadata.namespace.clone().unwrap());
+        let mut attached_process = pods.exec(&pod_name, psql_command, &attach_params).await.unwrap();
+        // attached_process.join();
```
ianstanton, line 185:
```suggestion
```

ianstanton (approved):

coredb-operator/src/controller.rs
ChuckHend, line 192: I think as its written, the only way `psql()` is going to return an error is if one of those `unwrap()` has an error. But, I think we'd want this to also return an error if there was anything logged to stderr? (maybe im wrong). So we could `return Err(result_stderr)`, OR `Ok(result_stdout)` AND before that, log both?

Mostly pseudo code of what I'm thinking.

```rust
use log::{info, warn};
info!(result_stdout);
if result_stderr.len() > 0 { 
   warn!(result_stderr) 
   Error(result_stderr)
} else {
  Ok(result_stdout)
}
```

coredb-operator/src/lib.rs
```diff
@@ -34,5 +34,6 @@ pub use metrics::Metrics;
 
 mod defaults;
 #[cfg(test)] pub mod fixtures;
+mod psql;
```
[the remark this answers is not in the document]
ChuckHend: I think we'd need this if we wanted to have a sub crate in a dir `src/psql`, then in that folder we'd have more stuff like lib.rs, sql.rs, somethingelse.rs. I don't see that in the PR though.

coredb-operator/tests/integration_tests.rs
ChuckHend: you're further along in Rust testing than I am :)

ChuckHend (approved):

coredb-operator/src/lib.rs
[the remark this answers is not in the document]
sjmiller609: whoops I added that and forgot. I was originally planning to put the psql in a new module

coredb-operator/src/controller.rs
sjmiller609: Good idea, I will change this part

coredb-operator/tests/integration_tests.rs
sjmiller609: I'll ping you if I get that part working, didn't yet

coredb-operator/src/lib.rs
```diff
@@ -1,3 +1,19 @@
+/// Expose all controller components used by main
```
sjmiller609, line 1: Darren mentioned these typically go at top of file

PR #43, opened 2023-01-13
[COR-131] CoreDB as a read replica to AWS RDS

sjmiller609:


sjmiller609:
looping back to this after setting up cli

PR #45, opened 2023-01-15
pgmq serialize + deserialize on enqueue / read

ChuckHend:
Handle serialization/deserialization on all reads/writes.

### DIFF
--- a/crates/pgmq/src/lib.rs
+++ b/crates/pgmq/src/lib.rs
@@ -1,18 +1,21 @@
-use sqlx::types::chrono::Utc;
-
+use serde::{Deserialize, Serialize};
 use sqlx::error::Error;
 use sqlx::postgres::{PgPoolOptions, PgRow};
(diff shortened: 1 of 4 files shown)


crates/pgmq/src/lib.rs
```diff
@@ -91,10 +97,3 @@ impl PGMQueue {
     //     // TODO: returns a struct
     // }
 }
-
-pub struct PGMQueueConfig {
-    pub url: String,
-    pub queue_name: String,
-    pub vt: u32,
-    pub delay: u32,
-}
```
ChuckHend, line 100: should have removed this a while ago.

ianstanton (approved):

sjmiller609 (approved):
Probably overkill for the moment, but could be worth considering using protobuf to decide how to store and deserialize messages.

https://github.com/tokio-rs/prost

PR #51, opened 2023-01-17
Setup CLI

sjmiller609:
Basic CLI setup PR


coredb-cli/src/commands/create.rs
```diff
@@ -0,0 +1,18 @@
+use super::{ResourceType, SubCommand};
```
ChuckHend, line 1: I _think_ alternatively, you could add a `lib.rs`. Then in `lib.rs`, `pub mod everything.rs`. Then through the project, including `main.rs` you can reference all the functions and structs like `coredb-cli::commands::{get, create}`

ChuckHend (approved):

coredb-cli/src/commands/create.rs
sjmiller609: Ok I will try this

PR #55, opened 2023-01-19
[COR-146] Add tests for CLI

sjmiller609:
Add a means to test the CLI by calling it with specific arguments, and add test execution in CI.

### DIFF
--- a/coredb-cli/tests/integration_tests.rs
+++ b/coredb-cli/tests/integration_tests.rs
@@ -1,0 +1,16 @@
+use assert_cmd::prelude::*; // Add methods on commands
+use predicates::prelude::*; // Used for writing assertions
+use std::process::Command; // Run programs
+
+const CARGO_BIN: &str = "coredb-cli";
+
+#[test]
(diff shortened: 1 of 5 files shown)


coredb-cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    let mut cmd = Command::cargo_bin(CARGO_BIN)?;
+    cmd.arg("create").arg("--dry-run").arg("db").arg("sample-db");
+    cmd.assert().stdout(predicate::str::contains("kind: CoreDB"));
+
+    let mut cmd = Command::cargo_bin(CARGO_BIN)?;
+    cmd.arg("create").arg("db").arg("sample-db").arg("--dry-run");
+    cmd.assert().stdout(predicate::str::contains("kind: CoreDB"));
+
+    Ok(())
```
sjmiller609, line 31: For the integration tests, I thought overkill and slow to try actually deploying it, since the operator's test will check actually deploying. Instead, we can test with the --dry-run flag and just check what yaml is generated.

PR #57, opened 2023-01-20
[COR-155] 'coredb install' to install operator

sjmiller609:
https://user-images.githubusercontent.com/7516283/213816129-dc8a6ec6-9871-4640-aa87-e2accff850f3.mov

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -85,6 +85,11 @@ jobs:
           prefix-key: "coredb-operator-functional-test"
           workspaces: |
             coredb-operator
+      - name: Check CRD is updated in the yaml directory
+        run: |
+          set -xe
+          cargo run --bin crdgen > yaml/crd.yaml
+          git diff --exit-code yaml/crd.yaml
       - name: Run functional / integration tests
         run: |
           set -xe
--- a/.github/workflows/publish_operator.yml
+++ b/.github/workflows/publish_operator.yml
@@ -1,4 +1,4 @@
-name: CoreDB Operator image build and upload
+name: CoreDB operator image build and upload
 
 on:
   push:
--- a/.github/workflows/publish_operator.yml
+++ b/.github/workflows/publish_operator.yml
@@ -1,0 +1,29 @@
+# TODO: replace with push to CoreDB Quay, and versioned pushes.
+# This is to enable demo 'coredb install' on the CLI.
+#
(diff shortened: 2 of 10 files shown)


.github/workflows/operator.yml
```diff
@@ -85,6 +85,11 @@ jobs:
           prefix-key: "coredb-operator-functional-test"
           workspaces: |
             coredb-operator
+      - name: Check CRD is updated in the yaml directory
+        run: |
+          set -xe
+          cargo run --bin crdgen > yaml/crd.yaml
+          git diff --exit-code yaml/crd.yaml
```
sjmiller609, line 92: this part will fail if CRD definition is changed but yaml directory not updated

.github/workflows/publish_operator.yml
```diff
@@ -0,0 +1,29 @@
+# TODO: replace with push to CoreDB Quay, and versioned pushes.
+# This is to enable demo 'coredb install' on the CLI.
+#
+on:
+  push:
+    branches:
+      - main
+      - cli
```
sjmiller609, line 8: todo remove 'cli' after I check this works

coredb-operator/Dockerfile
```diff
@@ -1,4 +1,17 @@
...
+
+FROM alpine:latest
+
+RUN adduser -D nonroot
+
+COPY --chown=nonroot:nonroot --from=builder target/x86_64-unknown-linux-musl/release/controller /app/controller
+COPY --chown=nonroot:nonroot --from=builder target/x86_64-unknown-linux-musl/release/crdgen /app/crdgen
+
+USER nonroot
+
```
sjmiller609, line 15: Although rust compiles to a binary, by default that binary has dependencies in the system like openssl and libc. If we build with MUSL we can avoid that - like clux did in the controller example.

coredb-operator/yaml/install.yaml
```diff
@@ -86,7 +86,7 @@ spec:
       serviceAccountName: coredb-controller
       containers:
       - name: coredb-controller
-        image: localhost:5001/controller:ac593af6da990c911652c6e35fb777f36498d6dc
+        image: sjmiller609/coredb-operator:latest
```
sjmiller609, line 89: replace this with quay when ian is back

ChuckHend (approved):

PR #59, opened 2023-01-23
Rename API

sjmiller609:
We have the API named unintentionally. It was just set as the default name in the kube-rs controller example. Instead, the API should be named to our domain.

### DIFF
--- a/coredb-cli/src/commands/create.rs
+++ b/coredb-cli/src/commands/create.rs
@@ -17,7 +17,7 @@ fn generate_yaml(resource_type: ResourceType, name: String) -> String {
     match resource_type {
         ResourceType::Db | ResourceType::Dbs => {
             let json_value = json!({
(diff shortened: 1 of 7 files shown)


coredb-cli/src/commands/create.rs
```diff
@@ -17,7 +17,7 @@ fn generate_yaml(resource_type: ResourceType, name: String) -> String {
     match resource_type {
         ResourceType::Db | ResourceType::Dbs => {
             let json_value = json!({
-                "apiVersion": "kube.rs/v1",
```
ianstanton, line 20: Thanks! @sjmiller609 what are your thoughts on starting with `coredb.io/v1alpha1`? We will likely have new API versions in the future.

sjmiller609: ok done

ianstanton (approved):

PR #60, opened 2023-01-24
[COR-159] Enable extensions provided in spec

ianstanton:
Add extension creation based on extensions specified in spec value `enabledExtensions`.
Example spec:
```
apiVersion: coredb.io/v1alpha1
kind: CoreDB
metadata:
  name: sample-coredb
spec:
  replicas: 1
  enabledExtensions:
    - pgmq
    - postgis
```

We'll iterate through this list and run `CREATE EXTENSION <extension-name>;` for each extension. This command will no-op if we've already created the extension.

### DIFF
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -48,6 +48,7 @@ pub struct CoreDBSpec {
     pub port: i32,
     #[serde(default = "defaults::default_uid")]
     pub uid: i32,
+    pub enabledExtensions: Vec<String>
 }
 
 /// The status object of `CoreDB`
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -49,7 +49,7 @@ pub struct CoreDBSpec {
     #[serde(default = "defaults::default_uid")]
     pub uid: i32,
     #[serde(default = "defaults::default_extensions")]
-    pub enabledExtensions: Vec<String>
+    pub enabledExtensions: Vec<String>,
 }
 
 /// The status object of `CoreDB`
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -22,7 +22,7 @@ use kube::{
     CustomResource, Resource,
 };
 
-use crate::{controller_util::create_extensions, secret::reconcile_secret};
+use crate::{extensions::create_extensions, secret::reconcile_secret};
 use k8s_openapi::api::core::v1::Pod;
 use kube::runtime::wait::{await_condition, conditions, Condition};
 use schemars::JsonSchema;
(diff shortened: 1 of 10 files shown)


coredb-operator/src/controller.rs
```diff
@@ -192,6 +229,21 @@ impl CoreDB {
     }
 }
 
+pub fn is_pod_ready() -> impl Condition<Pod> + 'static {
```
ianstanton, line 232: Moved this here from the test file because we're making use of it in the main controller loop.

coredb-operator/src/controller.rs
```diff
@@ -40,6 +41,7 @@ pub static COREDB_FINALIZER: &str = "coredbs.coredb.io";
 #[cfg_attr(test, derive(Default))]
 #[kube(kind = "CoreDB", group = "coredb.io", version = "v1alpha1", namespaced)]
 #[kube(status = "CoreDBStatus", shortname = "cdb")]
+#[allow(non_snake_case)]
```
ianstanton, line 44: Kubernetes yaml keys use camelCase.

coredb-operator/src/extensions.rs
```diff
@@ -0,0 +1,27 @@
...
+        debug!("Creating extension: {}", ext);
+        // this will no-op if we've already created the extension
+        let result = cdb
+            .psql(
+                format!("CREATE EXTENSION {};", ext),
+                "postgres".to_owned(),
+                client.clone(),
+            )
+            .await
+            .unwrap();
```
ChuckHend, line 23: Not immediately a concern, but we may want to evaluate this for sql injection vulnerability in the future. i.e. what happens if an extension named `; drop from pg_catalog` gets through?

ChuckHend (approved):

coredb-operator/src/controller.rs
sjmiller609: ah I see, it's rust that was having the issue with it

coredb-operator/src/controller.rs
sjmiller609: perfect

coredb-operator/src/extensions.rs
sjmiller609: I think in this case, we could validate it's a lower case alpha numerics and underscores

coredb-operator/tests/integration_tests.rs
```diff
@@ -146,13 +135,31 @@ mod test {
...
+        // Assert table 'customers' exists
         let result = coredb_resource
             .psql("\\dt".to_string(), "postgres".to_string(), client.clone())
             .await
             .unwrap();
         println!("{}", result.stdout.clone().unwrap());
         assert!(result.stdout.clone().unwrap().contains("customers"));
 
+        // TODO(ianstanton) we need to properly wait for 'postgis' extension to be created
+        thread::sleep(Duration::from_millis(500));
```
sjmiller609, line 148: I think this is possible to do with a 'condition', like we have in the case of is_pod_ready, but not necessary right away for a test

sjmiller609 (approved):
I think we should handle at least the input validation before merging

coredb-operator/src/extensions.rs
ChuckHend: I think what @sjmiller609 suggested works.

ChuckHend: Maybe longer term we bake something into `.psql()` to perform validation.

PR #63, opened 2023-01-24
[COR-167] Update default image to one with extensions

sjmiller609:


ChuckHend (approved):

coredb-operator/src/defaults.rs
```diff
@@ -11,5 +11,5 @@ pub fn default_port() -> i32 {
 }
 
 pub fn default_image() -> String {
-    "docker.io/postgres:15".to_owned()
+    "quay.io/coredb/postgres:2023.01.24".to_owned()
```
sjmiller609, line 14: we will eventually need tags corresponding to each supported postgres version

PR #65, opened 2023-01-25
[COR-166] Generate password and create secret

ianstanton:
TODO:
- [ ] Add test for secret creation

### DIFF
--- a/coredb-operator/src/secret.rs
+++ b/coredb-operator/src/secret.rs
@@ -1,0 +1,69 @@
+use crate::{Context, CoreDB, Error};
+use k8s_openapi::{api::core::v1::Secret, apimachinery::pkg::apis::meta::v1::ObjectMeta, ByteString};
+use kube::{
+    api::{Patch, PatchParams},
+    Api, Resource, ResourceExt,
+};
+use passwords::PasswordGenerator;
+use std::{collections::BTreeMap, sync::Arc};
+
+pub async fn reconcile_secret(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error> {
+    let client = ctx.client.clone();
+    let ns = cdb.namespace().unwrap();
+    let name = cdb.name_any();
+    let mut labels: BTreeMap<String, String> = BTreeMap::new();
+    let mut data: BTreeMap<String, ByteString> = BTreeMap::new();
+    let secret_api: Api<Secret> = Api::namespaced(client, &ns);
+    let oref = cdb.controller_owner_ref(&()).unwrap();
+    labels.insert("app".to_owned(), "coredb".to_owned());
+
+    // encode and insert user into secret data
+    let user = "postgres".to_owned();
+    let b64_user = b64_encode(user);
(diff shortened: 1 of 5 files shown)


ianstanton:
Opening up for review while I fix tests

coredb-operator/src/secret.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let ps = PatchParams::apply("cntrlr").force();
+    let _o = secret_api
+        .patch(&name, &ps, &Patch::Apply(&secret))
+        .await
+        .map_err(Error::KubeError)?;
+    Ok(())
+}
+
+fn b64_encode(string: &String) -> ByteString {
```
ChuckHend, line 81:
```suggestion
fn b64_encode(string: &str) -> ByteString {
```

coredb-operator/src/secret.rs
```diff
@@ -0,0 +1,100 @@
...
+        numbers: true,
+        lowercase_letters: true,
+        uppercase_letters: true,
+        symbols: false,
+        spaces: false,
+        exclude_similar_characters: false,
+        strict: true,
+    };
+    let password = pg.generate_one().unwrap();
+    password
```
ChuckHend, line 99:
```suggestion
    pg.generate_one().unwrap()
```

ChuckHend (approved):

coredb-operator/src/secret.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    // encode and insert host into secret data
+    let host = format!("{}.{}.svc.cluster.local", &name, &ns);
+    let b64_host = b64_encode(&host);
+    data.insert("host".to_owned(), b64_host);
+
+    // encode and insert uri into secret data
+    let uri = format!("postgresql://{}:{}@{}:{}", &user, &password, &host, &port);
+    let b64_uri = b64_encode(&uri);
+    data.insert("uri".to_owned(), b64_uri);
```
ChuckHend, line 58: We could probably move preparing of the `data` hashmap out to its own function, which would make this section easier to test. That function would look something like `fn prepare_data(name: &str, ns: &str) -> BTreeMap<String, ByteString>`. Maybe good for later.

ianstanton: Good idea, moved it to a separate fn!

coredb-operator/src/secret.rs
```diff
@@ -0,0 +1,106 @@
...
+
+fn b64_encode(string: &str) -> ByteString {
+    let bytes_vec = string.as_bytes().to_vec();
+    let byte_string = ByteString(bytes_vec);
+    byte_string
+}
+
+fn generate_password() -> String {
+    let pg = PasswordGenerator {
+        length: 8,
```
sjmiller609, line 96:
```suggestion
        length: 16,
```

sjmiller609: since will be hosting on public internet and without symbols to avoid encoding hassle

sjmiller609 (approved):
just one nit pick

PR #78, opened 2023-02-03
pgmq metadata

ChuckHend:
- adds two fields to each queue's table; 
   - `enqueued_at` - time the message hit the queue 
   - `read_ct` number of times the message has been read. if its >1 it usually means there's a problem somewhere in the application
- changed `queue.enqueue()` to `queue.send()`. This is easier to type and more in line with SQS and other messaging systems.
- added `pgmq_metadata` table. this contains one record for every queue and will be what we can query for `list_queues()`.  Currently only contains the name of the queues but will have other information later, like time created.
- added the `pgmq_<queue_name>_archive` table. It won't be in use in this PR, but we will be adding `queue.archive()` which will be the alternative to `queue.delete_message()`. Instead of deleting the message completely, it will move it to the archive table.


crates/pgmq/src/lib.rs
```diff
@@ -131,15 +131,17 @@ impl PGMQueue {
...
-        let create = query::create(queue_name);
-        let index: String = query::create_index(queue_name);
-        sqlx::query(&create).execute(&self.connection).await?;
-        sqlx::query(&index).execute(&self.connection).await?;
+        let mut tx = self.connection.begin().await?;
+        let setup = query::init_queue(queue_name);
+        for q in setup {
+            sqlx::query(&q).execute(&mut tx).await?;
+        }
+        tx.commit().await?;
```
ChuckHend, line 139: creating a queue now does a bunch of things, all are no-ops if the data already exists. metadata table, queue table, queue archive table, create index, insert metadata to queue metadata table

sjmiller609 (approved):

PR #79, opened 2023-02-04
[COR-175] Image build and upload

sjmiller609:
**This is the plan for our release automation:**

Image workflow (on commit to any branch): **This PR is only handling this part**

- Publish docker image with git hash as tag
- If main or release/** branch, also publish the version in cargo.toml, if the file is present
- If main branch, also publish ‘latest’

**Below items in following PRs**

Cargo workflow (on commit to main or release/** branch):

- publish cargo projects that are not already published (i.e. skip publish when version in Cargo.toml already exists on crates.io)

Github release workflow (on commit to release/** branch):

- github release for the major.minor version in the branch name, lowest unused patch number
- in release note, list versions of subcomponents

On PR to main or release/** branch:

- For changes in Cargo projects, fail if the version in Cargo.toml is already is published


.github/actions/release-version/action.yml
```diff
@@ -0,0 +1,142 @@
+name: 'Release version'
```
sjmiller609, line 1: This action is not yet being consumed. What it does it look at a release branch name, like release/0.1 and provides back the next release or prerelease version.

For example a branch like release/0.1, if there is already 0.1.0 and 0.1.1-rc.0 published should return the next version as 0.1.1 and the next prerelease as 0.1.1-rc.1

.github/actions/build-and-push-to-quay/action.yml
```diff
@@ -22,10 +22,18 @@ inputs:
     description: 'Should we tag with calender versioning?'
     required: false
     default: false
+  calver_suffix:
+    description: 'Optional suffix to the calendar version'
+    required: false
+    default: ""
```
sjmiller609, line 28: did not end up consuming this feature in a workflow, but might as well leave it in case we want this flag later.

.github/workflows/publish_operator.yml
```diff
@@ -1,26 +0,0 @@
-name: CoreDB operator image build and upload
```
sjmiller609, line 1: this change consolidates all image build workflows.

ChuckHend (approved):

PR #83, opened 2023-02-05
[COR-175] Check that updated crates have their version updated

sjmiller609:


.github/actions/publish-crate/action.yml
```diff
@@ -0,0 +1,71 @@
...
+        echo "The list of known versions:"
+        echo $PUBLISHED_VERSIONS
+
+        # Set --dry-run flag, if configured
+        DRY_RUN=""
+        if [ "${{ inputs.dry-run }}" == "true" ]; then
+          DRY_RUN="--dry-run"
+        fi
+
+        cargo publish ${DRY_RUN} --token ${{ inputs.cargo-registry-token }}
```
sjmiller609, line 71: --dry-run flag does not fail when the version is already published, so that is why we have another step before this one that can conditionally fail if the version is already published.

coredb-cli/README.md
```diff
@@ -1,3 +1,4 @@
 # CoreDB CLI
 
+
```
sjmiller609, line 3: nothing changed, just to check on the publishing workflow

ChuckHend (approved):

ChuckHend (approved):

PR #84, opened 2023-02-06
[COR-145] Log SQL Statements at Debug Level

ianstanton:
By default, sqlx logs SQL statements at info level. This can become very noisy in application logs. This PR allows us to configure the log level by adding `PgConnectOptions`. 

For now, we'll set `options.log_statements(LevelFilter::Debug);` and allow for user configurable log level in a follow-up PR.


crates/pgmq/tests/integration_test.rs
```diff
@@ -329,7 +329,7 @@ async fn test_database_error_modes() {
     // we expect a database error
     match queue {
         Err(e) => {
-            if let pgmq::errors::PgmqError::DatabaseError { .. } = e {
+            if let pgmq::errors::PgmqError::UrlParsingError { .. } = e {
```
ChuckHend, line 332: can you add a test that covers the `DatabaseError` case? I think it would look something like:
- provided a VALID database connection string, but with either wrong host or password, `pgmq::errors::PgmqError::DatabaseError` is returned.

ChuckHend: Keep the existing test that makes sure `DatabaseError` happens when its the wrong, but valid, connection string.

Add new test that validates that a poorly formed connection string returns `UrlParsingError`

ChuckHend (approved):

crates/pgmq/tests/integration_test.rs
ianstanton: Yep, I'll add that

PR #90, opened 2023-02-07
Build packages for postgres extensions

sjmiller609:
I'm looking at how postgres ML is doing packaging:

They are building with pgx inside docker and then copying a `.deb` package out from the container, which can be installed on ubuntu.

- End user doc: https://postgresml.org/user_guides/setup/v2/installation/#install-the-extension
- script to build in docker https://github.com/postgresml/postgresml/blob/master/pgml-extension/build_extension.sh
- dockerfile where they do the build https://github.com/postgresml/postgresml/blob/master/pgml-extension/Dockerfile
- script to create a .deb package after cargo pgx package https://github.com/postgresml/postgresml/blob/master/pgml-extension/docker/build_ubuntu_deb.sh

At the moment, we do not have an APT repository. I think we ought to build and upload our extensions to an APT repository, then install from there into our postgres image, that way our extensions and other extensions are being installed in the same way. For this PR, we are just building the extension before doing the docker build of postgres, then using COPY in the dockerfile to copy the .deb into our postgres image and install it.


.github/actions/build-extension-package/Dockerfile
```diff
@@ -0,0 +1,54 @@
+ARG UBUNTU_VERSION=22.04
```
sjmiller609, line 1: This is imitating this dockerfile, less the not applicable parts 

https://github.com/postgresml/postgresml/blob/master/pgml-extension/Dockerfile

.github/workflows/cargo.yml
```diff
@@ -32,52 +32,52 @@ jobs:
...
-        with:
-          working-directory: ${{ matrix.path }}
-      - name: Publish or validate
-        uses: ./.github/actions/publish-crate
-        with:
-          working-directory: ${{ matrix.path }}
-          dry-run: ${{ steps.cargo_flags.outputs.dry_run }}
-          fail-if-version-published: ${{ steps.cargo_flags.outputs.fail_if_version_published }}
-          cargo-registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+  # cargo_publish:
```
sjmiller609, line 35: TODO uncomment before merge

.github/workflows/pg_smq_ext.yml
```diff
@@ -20,55 +20,69 @@ on:
...
-          cache-directories: |
-            /home/runner/.pgx
-      - uses: ./.github/actions/pgx-init
-        with:
-          working-directory: extensions/pgx_pgmq
-      - name: Cargo format
-        run: cargo +nightly fmt --all --check
-      - name: Clippy
-        run: cargo clippy
+  # lint:
```
sjmiller609, line 23: TODO uncomment before merge

extensions/pgx_pgmq/Cargo.toml
```diff
@@ -25,7 +25,7 @@ pg_test = []
 [dependencies]
 pgx = "0.7.1"
 serde = "1.0.152"
-pgmq = { path = "../../crates/pgmq", version = ">=0.4.0, <1.0.0" }
+pgmq = ">=0.4.0, <1.0.0"
```
sjmiller609, line 28: undecided if this is a good idea - it's because if we are building the extension with Docker, then there is some hassle to use the relative pathing

extensions/pgx_pgmq/build-extension.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
sjmiller609, line 1: similar to this script

https://github.com/postgresml/postgresml/blob/master/pgml-extension/build_extension.sh

sjmiller609: It uses a container to build the extension, and outputs to the local filesystem the `.deb` package

extensions/pgx_pgmq/Cargo.toml
ChuckHend: Ah ya, good point.

.github/actions/build-extension-package/Dockerfile
sjmiller609: The purpose of this file is to build an .deb (package) file for ubuntu with a PGX extension in it

postgres/Dockerfile
```diff
@@ -1,7 +1,51 @@
-FROM postgres:15.1
+FROM ubuntu:22.04
```
sjmiller609, line 1: Since we built on ubuntu, the shared libraries don't work unless we also run on Ubuntu. We kind of have to pick one distro to support first, might as well be ubuntu. That's what Postgres ML is doing.

postgres/docker-entrypoint.sh
```diff
@@ -0,0 +1,340 @@
+#!/usr/bin/env bash
```
sjmiller609, line 1: This is copy/pasted from here https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh

.github/actions/build-and-push-to-quay/action.yml
```diff
@@ -85,6 +85,16 @@ runs:
...
+    - name: Run pre-build hooks
+      shell: bash
+      run: |
+        cd ${{ inputs.docker_directory }}
+        if [[ -f pre-build-hook.sh ]]; then
+          echo "detected pre-build hook, running"
+          /bin/bash pre-build-hook.sh
+        else
+          echo "no pre build hook detected"
+        fi
```
sjmiller609, line 97: To get the extension installed into the postgres container, we have a script that builds the extension, and puts the package into a directory "extensions" in the postgres directory before doing the build of the postgres image.

I think we ought to instead publish our packages and just pull them in in the docker build over the internet from our repository instead of doing this

.github/actions/build-extension-package/docker/build_ubuntu_deb.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
sjmiller609, line 1: similar to https://github.com/postgresml/postgresml/blob/master/pgml-extension/docker/build_ubuntu_deb.sh

extensions/pgx_pgmq/Cargo.toml
sjmiller609: TODO: review this

.github/actions/build-and-push-to-quay/action.yml
ChuckHend: publish the pgx extension as a crate, then pull from crates.io and install?

as opposed to cloning the repo, then pushing to crates.io and quay.io at same time (or in sequence)?

.github/actions/build-extension-package/Dockerfile
ChuckHend: will all of our CoreDB postgres images be ubuntu based then, or just the images that contain extensions which we author?

.github/actions/build-and-push-to-quay/action.yml
sjmiller609: I mean publish to an APT repository the .deb package. Seems to me like that's the best way to package and distribute extensions.

.github/actions/build-extension-package/Dockerfile
sjmiller609: I think we have to pick one distro to do first. For example, postgres ML is only doing ubuntu or inside a container (or build from source): https://postgresml.org/user_guides/setup/v2/installation/#install-the-extension

extensions/pgx_pgmq/Cargo.toml
sjmiller609: It's possible to copy the crate before building the container. Which do you think? I don't mind either way.

extensions/pgx_pgmq/README.md
```diff
@@ -180,7 +180,7 @@ TODO:
 Delete a message with id `1` from queue named `my_queue`.
 ```sql
 pgmq=# select pgmq_delete('my_queue', 1);
- pgmq_delete 
+ pgmq_delete
```
sjmiller609, line 183: my editor is automatically removing trailing whitespaces

.github/actions/build-and-push-to-quay/action.yml
sjmiller609: Like I mean i'm imagining our workflow will build and publish to our APT repository all our extensions (and I think we don't need to publish extensions to crates.io). So that's another artifact type.

extensions/pgx_pgmq/Cargo.toml
ChuckHend: lets point it to crates.io like you have it

postgres/docker-entrypoint.sh
```diff
@@ -0,0 +1,340 @@
...
+# check to see if this file is being run or sourced from another script
+_is_sourced() {
+	# https://unix.stackexchange.com/a/215279
+	[ "${#FUNCNAME[@]}" -ge 2 ] \
+		&& [ "${FUNCNAME[0]}" = '_is_sourced' ] \
+		&& [ "${FUNCNAME[1]}" = 'source' ]
+}
+
+# used to create initial postgres directories and if run as root, ensure ownership to the "postgres" user
+docker_create_db_directories() {
```
ChuckHend, line 36: where are we setting any values that need to land in `postgresql.conf`? For example, `shared_preload_libraries = 'pg_cron'
`

ChuckHend (approved):

ianstanton (approved):

PR #91, opened 2023-02-07
[COR-231] Add read_batch() to `pgmq`

ianstanton:
Add function `read_batch()` to pgmq.

Related to: https://linear.app/coredb/issue/COR-231/read-batch-messages

### DIFF
--- a/crates/pgmq/Cargo.toml
+++ b/crates/pgmq/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.4.0"
+version = "0.4.1-alpha.1"
 edition = "2021"
 authors = ["CoreDB.io"]
 description = "A message queue for Rust applications. The only external dependency is a Postgres database."
--- a/crates/pgmq/Cargo.toml
+++ b/crates/pgmq/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.4.1-alpha.1"
+version = "0.5.0"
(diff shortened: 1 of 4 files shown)


crates/pgmq/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.4.0"
+version = "0.4.1-alpha.1"
```
ianstanton, line 3: Not sure if we need to bump now or if we should wait for all batch features.

crates/pgmq/tests/integration_test.rs
```diff
@@ -179,6 +179,38 @@ async fn test_fifo() {
...
+    let batch = queue
+        .read_batch::<Value>(&test_queue, Some(&vt), &num_msgs)
+        .await
+        .unwrap()
+        .unwrap();
+
+    for (i, message) in batch.iter().enumerate() {
+        let index = i + 1;
+        assert_eq!(message.msg_id.to_string(), index.to_string());
+    }
```
ChuckHend, line 211: can you add a test that validates the visibility timeout is getting set appropriately? something like:

```rust
    let msg = queue.read::<Value>(&test_queue, Some(&vt)).await.unwrap();
    assert!(msg.is_none());
    let num_rows = rowcount(&test_queue, &queue.connection).await;
    // no messages on read but still 3 records on the table
    assert_eq!(num_rows, 3);
    ```

ChuckHend (approved):

PR #93, opened 2023-02-08
Fix pgx_pgmq read

ianstanton:
Read function parameters recently changed. Updating in `pgx_pgmq` extension.


ianstanton:
Need to investigate failures after merging in main (CI was green)

ianstanton:
`pgmq` dependency was reverted to `0.4.0`

sjmiller609 (approved):
I would increment the version of pgx_pgmq to 0.0.4 or 0.5.1

extensions/pgx_pgmq/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pgx_pgmq"
-version = "0.0.3-alpha.1"
+version = "0.5.1"
```
ChuckHend, line 3: Are we going to tie these versions together? I'd expect the crates development to slow over time and the extension to pick up.

ianstanton: Let's go with `0.0.4` in that case

ChuckHend (approved):

PR #111, opened 2023-02-14
[COR-247] Add postgres_exporter role

sjmiller609:
This PR creates a role postgres_exporter like how the postgres exporter's docs indicate to do. There will be a subsequent PR to serve these metrics on the pod using pg exporter.

### DIFF
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -48,6 +48,8 @@ pub static COREDB_FINALIZER: &str = "coredbs.coredb.io";
 pub struct CoreDBSpec {
     #[serde(default = "defaults::default_replicas")]
     pub replicas: i32,
+    #[serde(default = "defaults::default_postgres_exporter_role_enabled")]
+    pub postgres_exporter_role_enabled: bool,
     #[serde(default = "defaults::default_image")]
     pub image: String,
     #[serde(default = "defaults::default_port")]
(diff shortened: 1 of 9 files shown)


sjmiller609:
hmm for some reason the function tests failed, but they passed locally. Looking into it.

coredb-operator/yaml/crd.yaml
```diff
@@ -31,12 +31,15 @@ spec:
                   type: string
                 type: array
               image:
-                default: quay.io/coredb/postgres:2023.01.24
+                default: quay.io/coredb/postgres:c03124e
```
sjmiller609, line 34: It was failing because the default user of this container is postgres, so I fixed by updating the init container to run as root. It did not fail locally because I forgot to locally apply the CRD again.

coredb-operator/src/controller.rs
```diff
@@ -45,6 +48,8 @@ pub static COREDB_FINALIZER: &str = "coredbs.coredb.io";
 pub struct CoreDBSpec {
     #[serde(default = "defaults::default_replicas")]
     pub replicas: i32,
+    #[serde(default = "defaults::default_postgres_exporter_enabled")]
+    pub postgresExporterEnabled: bool,
```
ianstanton, line 52: Would we ever want to configure the `postgresExporter`? If yes, we could do something like:
```
spec:
  postgresExporter:
    enabled: true
    config: value
```

coredb-operator/tests/integration_tests.rs
```diff
@@ -145,7 +145,7 @@ mod test {
         assert!(result.stdout.clone().unwrap().contains("customers"));
 
         // TODO(ianstanton) we need to properly wait for 'postgis' extension to be created
-        thread::sleep(Duration::from_millis(500));
+        thread::sleep(Duration::from_millis(5000));
```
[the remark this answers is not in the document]
ianstanton: Yep, sleep is good for now! We'll add the watches as the project grows :)

ianstanton (approved):
LGTM, just one comment on spec value

PR #116, opened 2023-02-15
wip partman

ChuckHend:


sjmiller609 (approved):

extensions/pgmq/src/lib.rs
```diff
@@ -303,17 +321,73 @@ mod tests {
                 .expect("SQL select failed");
         assert_eq!(init_count.unwrap(), 0);
     }
+
```
ianstanton, line 324: Nice tests here :)

ianstanton (approved):

extensions/pgmq/src/lib.rs
ChuckHend: yes, thank you for them :)

PR #117, opened 2023-02-15
Add postgres exporter sidecar and metrics service

sjmiller609:


coredb-operator/src/controller.rs
```diff
@@ -146,8 +148,10 @@ impl CoreDB {
             debug!("Did not find primary pod");
             return Ok(Action::requeue(Duration::from_secs(1)));
         }
-        if !is_pod_ready().matches_object(Some(&primary_pod.unwrap())) {
-            debug!("Did not find primary pod is ready");
+        let primary_pod = primary_pod.unwrap();
+
+        if !is_postgres_ready().matches_object(Some(&primary_pod)) {
```
sjmiller609, line 153: we wait just for postgres to be ready, not the whole pod, before creating the postgres exporter role

coredb-operator/src/psql.rs
```diff
@@ -45,7 +45,7 @@ impl PsqlCommand {
 
     pub async fn execute(&self) -> Result<PsqlOutput, kube::Error> {
         let attach_params = AttachParams {
-            container: None,
+            container: Some("postgres".to_string()),
```
sjmiller609, line 48: now we have two containers in the pod, so add the container name to psql into

coredb-operator/src/statefulset.rs
```diff
@@ -53,6 +56,74 @@ pub fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
             ..VolumeMount::default()
         },
     ]);
+    let mut containers = vec![
```
sjmiller609, line 59: move containers into separate variable in order to enable conditionally appending the sidecar

ianstanton (approved):
@sjmiller609 LGTM! Thoughts on following up with a simple 'scrape metrics' test? We could argue this is not the operator's responsibility, but I think it's good to test the metrics are served as expected.

PR #118, opened 2023-02-15
[COR-255] CoreDB operator: don't block while waiting for pod to be ready

sjmiller609:
Instead, return to the reconciler loop with a short wait.

### DIFF
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -27,7 +27,7 @@ use crate::{
     secret::reconcile_secret,
 };
 use k8s_openapi::api::core::v1::{Namespace, Pod};
-use kube::runtime::wait::{await_condition, conditions, Condition};
+use kube::runtime::wait::Condition;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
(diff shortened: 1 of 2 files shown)


coredb-operator/src/controller.rs
```diff
@@ -144,30 +141,15 @@ impl CoreDB {
...
-        ));
-        debug!("Found pod ready: {}", pod_name);
+        let primary_pod = self.primary_pod(ctx.client.clone()).await;
+        if primary_pod.is_err() {
+            debug!("Did not find primary pod");
+            return Ok(Action::requeue(Duration::from_secs(1)));
+        }
+        if !is_pod_ready().matches_object(Some(&primary_pod.unwrap())) {
+            debug!("Did not find primary pod is ready");
+            return Ok(Action::requeue(Duration::from_secs(1)));
```
sjmiller609, line 151: This is 1 second wait hardcoded, but I think it ought to be an exponential backoff, up to X seconds maximum backoff time. However, I do not know how to get the previous queue duration yet.

sjmiller609: https://cloud.redhat.com/blog/kubernetes-operators-best-practices

sjmiller609: I think they are writing to 'status', that could be a future improvement to prevent checking on stuck pods every second

ianstanton (approved):
Boom!

PR #122, opened 2023-02-16
[COR-205] Add `resources` to CoreDB spec

ianstanton:
Add `resources` spec value. Use this value to set resource values on postgres container. Also add default values.


ianstanton:
`test::functional_test_delete_namespace` seems to be failing in CI. Passes locally...

coredb-operator/yaml/crd.yaml
```diff
@@ -47,6 +47,91 @@ spec:
...
+
+                        Non-canonical values will still parse as long as they are well formed, but will be re-emitted in their canonical form. (So always use canonical form, or don't diff.)
+
+                        This format is intended to make it difficult to use these numbers without writing some sort of special handling code in the hopes that that will cause implementors to also use a fixed point implementation.
+                      type: string
+                    description: 'Limits describes the maximum amount of compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
+                    type: object
+                  requests:
+                    additionalProperties:
+                      description: |-
```
ianstanton, line 99: These descriptions can cause unnecessary bloat in our CRD. Go's `controller-gen` has a `maxDescLen` flag that you can set to `0` to disable these descriptions. There doesn't seem to be an equivalent in `kube-derive` at the moment.

It's fine to include them at the moment, but wanted to call this out in case we need to trim things down in the future.

coredb-operator/src/defaults.rs
```diff
@@ -1,7 +1,25 @@
...
 }
 
+pub fn default_resources() -> ResourceRequirements {
+    let limits: BTreeMap<String, Quantity> = BTreeMap::from([
+        ("cpu".to_owned(), Quantity("2".to_string())),
+        ("memory".to_owned(), Quantity("2Gi".to_string())),
+    ]);
+    let requests: BTreeMap<String, Quantity> = BTreeMap::from([
+        ("cpu".to_owned(), Quantity("500m".to_string())),
+        ("memory".to_owned(), Quantity("512Mi".to_string())),
```
ianstanton, line 15: @sjmiller609 let me know your thoughts on these default values

ChuckHend (approved):

PR #124, opened 2023-02-17
pgmq input validation

ChuckHend:
Don't allow malformed values into SQL statements. Panic anytime there's invalid SQL.

### DIFF
--- a/crates/pgmq/src/query.rs
+++ b/crates/pgmq/src/query.rs
@@ -244,7 +244,7 @@ pub fn pop(name: &str) -> String {
 }
 
 /// panics if input is valid queue. otherwise does nothing.
-fn check_input(input: &str) {
+pub fn check_input(input: &str) {
     lazy_static! {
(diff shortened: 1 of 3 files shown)


crates/pgmq/src/query.rs
```diff
@@ -141,18 +155,9 @@ pub fn enqueue(name: &str, messages: &[serde_json::Value]) -> String {
...
-pub fn enqueue_str(name: &str, message: &str) -> String {
-    // TOOO: vt should be now() + delay
-    format!(
-        "
-        INSERT INTO {TABLE_PREFIX}_{name} (vt, message)
-        VALUES (now() at time zone 'utc', '{message}'::json)
-        RETURNING msg_id;
-        "
-    )
-}
```
ChuckHend, line 153: this was only used in the extension, so its being moved over there.

ianstanton (approved):
:boom:

PR #130, opened 2023-02-17
impl archive in pgmq extension

ChuckHend:
- archive has been in the crate for a while, lets implement it in the extension too
- follow-up to https://github.com/CoreDB-io/coredb/pull/125 - returns error enums instead of panicing


ianstanton (approved):
LGTM!

PR #131, opened 2023-02-20
Adds Ruby example in parity with the Python example

mkrisher:
This PR adds a Ruby example that is at parity with the existing Python example. It creates the extension, creates a queue, publishes a message, reads the message, deletes the message, and shows that reading messages off the queue can return an empty result.

Many Rails apps use Postgres as their data store, but also implement a message queue service for running their background jobs (via Redis and Sidekiq most often). This example shows the existing data store could be used potentially. It's relatively easy to conceive of a Ruby gem that would provide a simple DSL for managing queues and messages.

### DIFF
--- a/extensions/pgmq/examples/ruby.rb
+++ b/extensions/pgmq/examples/ruby.rb
@@ -1,0 +1,83 @@
+#!/usr/bin/env ruby
+
+# NOTES: to run this file, chmod 755 and then ./ruby.rb
+# this file demonstrates how to create a message queue, list the queues, post a message, read a
+# message, delete a message, and determine if a queue is empty
+#
+# Many Rails apps will use a queue such as Redis or AWS SQS or RabbitMQ, while using postgres as
+# the data store. Using pgmq eliminiates the need to support another service to queue messages for
+# a pattern like pub/sub or background jobs. A gem could be introduced with a simple DSL for all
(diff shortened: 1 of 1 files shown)


ryw:
Thanks @mkrisher :)

ChuckHend (approved):
Awesome @mkrisher! Sometime soon we want to get client libraries for the most popular languages, which includes Ruby. Any interest in helping with that?

mkrisher:
@ChuckHend I'd be more than happy to help. I think a lot of the Ruby community would utilize the extension so they don't have to run something in addition to Postgres. Are you thinking a library just for the extension, or coredb as a whole?

ChuckHend:
My thought was that it would be a pretty lightweight library that wraps up the sql statements for pgmq and simplify serialization. The library would be most useful when the extension is already installed on postgres. I was thinking it wouls be almost like an ORM but for the queues only (create, send, read, delete, archive).

Instead of writing `select * from pgmq_read()` the library would expose just a `read()` function and abstract the sql parts.

mkrisher:
gotcha, that was exactly what I was thinking as I worked on this Ruby example

ChuckHend:
We're also in Discord if you ever want to hear from the team 😄 
https://discord.gg/7bGYA9NPux

PR #132, opened 2023-02-21
Pgmq/clone

ChuckHend:


crates/pgmq/src/lib.rs
```diff
@@ -184,7 +184,7 @@ pub struct Message<T = serde_json::Value> {
 }
 
 /// Main controller for interacting with a queue.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
ianstanton, line 187: What does this do?

ChuckHend: implements the Clone trait on the struct. So that you can do stuff like:

```rust
let q = PgmQueue();
q.clone()
```

PR #134, opened 2023-02-22
Storybook

lukemoderwell:
Basic component library setup with Storybook. Includes `.scss` support:

Screenshot 2023-02-22 at 8 33 55 AM


ChuckHend (approved):

pgUI/pgUI/stories/Button.stories.ts
```diff
@@ -0,0 +1,23 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+// import { Button } from './Button';
+import Button from '../components/Button';
```
lukemoderwell, line 4: Not a very sophisticated example but here is a simple button. We can start adding components that are more advanced like the InstanceCard or even a certain extension UI view as we go.

PR #137, opened 2023-02-28
Problem: use of docker container in integration tests

yrashk:
This requires a correct set up and can also lead to some resource sharing issues.

Solution: make integration tests spin postgres up

This does *not* yet address the same issue for rustdoc so the existing Docker harness is left in place.


yrashk:
Note: this PR hits some (yet unknown) issue in CI (works fine on my end) with tests timing out over what looks like a corrupt archive of Postgres but in fact has something to do with async mutex locking not doing its job well. 

Pending further investigation time-permitting.

PR #147, opened 2023-03-06
drop extension / update crd

ChuckHend:
Operator creates/drops postgres extensions.

Updated the CRD for extensions to be a list of objects, instead of a list of strings. Each object has as attribute: `enabled`. Operator will use that attribute to switch its behavior.

### DIFF
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -194,7 +194,7 @@ impl CoreDB {
         }
 
         manage_extensions(self, ctx.clone()).await.expect(&format!(
-            "Error creating extensions on CoreDB {}",
+            "Error updating extensions on CoreDB {}",
             self.metadata.name.clone().unwrap()
         ));
 
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -23,7 +23,7 @@ use kube::{
 };
 
 use crate::{
-    extensions::create_extensions, postgres_exporter_role::create_postgres_exporter_role,
+    extensions::manage_extensions, postgres_exporter_role::create_postgres_exporter_role,
     secret::reconcile_secret,
 };
 use k8s_openapi::{
@@ -193,7 +193,7 @@ impl CoreDB {
             return Ok(Action::requeue(Duration::from_secs(1)));
         }
 
-        create_extensions(self, ctx.clone()).await.expect(&format!(
+        manage_extensions(self, ctx.clone()).await.expect(&format!(
             "Error creating extensions on CoreDB {}",
             self.metadata.name.clone().unwrap()
         ));
--- a/coredb-operator/src/controller.rs
+++ b/coredb-operator/src/controller.rs
@@ -66,7 +66,7 @@ pub struct CoreDBSpec {
     #[serde(default = "defaults::default_uid")]
     pub uid: i32,
     #[serde(default = "defaults::default_extensions")]
-    pub enabledExtensions: Vec<String>,
+    pub extensions: Vec<Extension>,
 }
 
 /// The status object of `CoreDB`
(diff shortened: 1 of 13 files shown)


coredb-operator/tests/integration_tests.rs
```diff
@@ -255,6 +255,42 @@ mod test {
...
+        // Apply crd with extension disabled
+        let params = PatchParams::apply("coredb-integration-test");
+        let patch = Patch::Apply(&coredb_json);
+        let coredb_resource = coredbs.patch(name, &params, &patch).await.unwrap();
+
+        // give it time to drop
+        thread::sleep(Duration::from_millis(5000));
+
+        // Assert extension no longer created
+        // Assert extension 'postgis' was created
```
ChuckHend, line 281:
```suggestion
```

coredb-operator/src/controller.rs
```diff
@@ -86,6 +86,25 @@ pub struct Context {
...
+impl Default for Extension {
+    fn default() -> Self {
+        Extension {
+            name: "pg_stat_statements".to_owned(),
+            version: "1.9".to_owned(),
+            enabled: true,
+            schema: "postgres".to_owned(),
+        }
+    }
+}
```
ianstanton, line 106:
```suggestion
```
I think the default should be an empty vec.

coredb-operator/src/defaults.rs
```diff
@@ -44,6 +46,6 @@ pub fn default_postgres_exporter_image() -> String {
     "quay.io/prometheuscommunity/postgres-exporter:v0.11.1".to_owned()
 }
 
-pub fn default_extensions() -> Vec<String> {
-    vec![]
+pub fn default_extensions() -> Vec<Extension> {
+    vec![Extension::default()]
```
ianstanton, line 50:
```suggestion
    vec![]
```
Same as above, I think we should default to empty `Vec<Extension>`.

coredb-operator/src/extensions.rs
```diff
@@ -1,32 +1,50 @@
...
-                )
-                .await
-                .unwrap();
-            debug!("Result: {}", result.stdout.clone().unwrap());
+            if ext.enabled {
+                info!("Creating extension: {}", ext_name);
+                // this will no-op if we've already created the extension
+                let result = cdb
+                    .psql(
+                        format!("CREATE EXTENSION IF NOT EXISTS {ext_name};"),
```
ianstanton, line 28: :ok_hand:

coredb-operator/src/extensions.rs
```diff
@@ -1,32 +1,50 @@
...
+                        "postgres".to_owned(),
+                        client.clone(),
+                    )
+                    .await
+                    .unwrap();
+                debug!("Result: {}", result.stdout.clone().unwrap());
+            } else {
+                info!("Dropping extension: {}", ext_name);
+                // this will no-op if we've already created the extension
```
ianstanton, line 37:
```suggestion
```

ianstanton (changes requested):

coredb-operator/src/extensions.rs
ChuckHend: sure, this was left over from previous implementation!

coredb-operator/src/controller.rs
ianstanton: What's an example case for using this? I could see us using it as a default CRD value if we wanted this extension enabled whenever the user provides nothing in `spec.extensions` (like you had previously configured), but we want that to be empty in that case.

ChuckHend: I use `::Default()` a lot in tests or wherever I want to construct a default set of values for struct.

ianstanton: :+1: seems reasonable!

ianstanton (approved):

coredb-operator/src/controller.rs
ChuckHend: `default_extensions` is an empty vec regardless of implementing Default on the Extension

sjmiller609 (approved):

PR #150, opened 2023-03-07
Use trunk for building PGMQ

sjmiller609:
Also, writes a script to serve in place of 'trunk install'.

Since we are still doing PGMQ build, it's not as fast as it will be when we can just "trunk install" the precompiled version from a registry.

Also this just removes a lot of code, the logic for building a PGX extension into a package.


postgres/pre-build-hook.sh
```diff
@@ -4,12 +4,11 @@ set -xe
...
 
-cd ${GIT_ROOT}/extensions/pgmq
-/bin/bash build-extension.sh
+cargo install --version 0.0.1-alpha.3 pg-trunk
 
-cd ${GIT_ROOT}/postgres
 mkdir extensions || true
 
-cp ${GIT_ROOT}/extensions/pgmq/*.deb ./extensions/
+trunk build --path ${GIT_ROOT}/extensions/pgmq/ --output-path ${THIS_DIR}/extensions
```
sjmiller609, line 14: "trunk build" replaces the code that was deleted in this PR

ChuckHend (approved):

ianstanton (approved):

PR #151, opened 2023-03-08
Extension per database

ChuckHend:
Manage extension in each database separately, and install them in a schema.

### DIFF
--- a/coredb-operator/src/defaults.rs
+++ b/coredb-operator/src/defaults.rs
@@ -49,3 +49,11 @@ pub fn default_postgres_exporter_image() -> String {
 pub fn default_extensions() -> Vec<Extension> {
     vec![]
 }
+
+pub fn default_database() -> String {
+    "postrgres".to_owned()
+}
+
+pub fn default_schema() -> String {
+    "public".to_owned()
+}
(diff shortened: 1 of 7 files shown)


sjmiller609 (approved):

coredb-operator/src/defaults.rs
```diff
@@ -49,3 +49,11 @@ pub fn default_postgres_exporter_image() -> String {
...
     vec![]
 }
+
+pub fn default_database() -> String {
+    "postrgres".to_owned()
+}
+
+pub fn default_schema() -> String {
+    "public".to_owned()
+}
```
ianstanton, line 59: :+1:

coredb-operator/yaml/sample-coredb.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
 spec:
   replicas: 1
   extensions:
-    - name: "pg_stat_statements"
-      enabled: true
-      version: "1.9"
-      schema: "postgres"
+    - name: pg_stat_statements
+      locations: 
+        - enabled: true
```
ianstanton, line 10: Nitpick - I like including the extra spec values such as `database`, `schema`, `version` in the sample. It could help a newcomer better understand what the `locations` value is for.

ianstanton (approved):

coredb-operator/yaml/sample-coredb.yaml
ChuckHend: I just took these out per @sjmiller609 request

ChuckHend: brought them back in!

PR #153, opened 2023-03-08
Handle expanding storage in CoreDB operator

shahadarsh:
* Add validation for StorageClass (allowVolumeExpansion property) -->  We should figure out a way to specify which StorageClass to use. Maybe add a property in `CoreDB` CRD?
* What if the storage is reduced? How to handle it? - Maybe add a validation for this?
* Better error handling --> There is a separate story for this

Done:
* Tested changes by deploying it to an env on AWS and it works as expected. Had to make [this](https://github.com/CoreDB-io/cloudknit-config/commit/e7a926e6c6d55a29431fd13b5e29b81e9f1dd1d2) change to add csi-resizer/.


coredb-operator/src/statefulset.rs
```diff
@@ -228,15 +228,78 @@ pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error>
...
+        .await
+        .map_err(Error::KubeError);
+}
+
+async fn update_pvc(client: Client, sts: &StatefulSet, cdb: &CoreDB) {
+    let pvc_api: Api<PersistentVolumeClaim> =
+        Api::namespaced(client, &sts.clone().metadata.namespace.unwrap());
+
+    let sts_name = sts.clone().metadata.name.unwrap();
+    let pvc_name = format!("data-{}-0", sts_name);
```
shahadarsh, line 273: Maybe there is a better way to query the PVC (i.e. without the "-0"). Need to take a look at it

sjmiller609 (approved):

ChuckHend (approved):

PR #178, opened 2023-03-23
Clean up CI failures after `trunk` migration

ianstanton:
There were a handful of CI failures after moving the `trunk` repositories. This is an attempt to resolve those failures.


ianstanton:
This may take a few tries, but this PR should resolve a couple of failures :)

ChuckHend (approved):

PR #183, opened 2023-03-26
Rusty/tinyoptimization

russellkemmit:
- Update send function to use an array for storing the msg inside of an iterable before it is sent to the enqueue function. By using an array we can take advantage of the tiny memory win because the array is put onto the stack as the size of the array is known at the time of its creation, whereas the vector is placed on the heap due to its dynamic nature. We also get to remove one line of code :)!
- Bump patch version.


ChuckHend:
Awesome!

ChuckHend (approved):

PR #190, opened 2023-03-30
pg permissions for extension dirs

ChuckHend:


postgres/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "coredb"
-version = "15.1.0-coredb.11"
+version = "15.2-coredb.1"
```
ChuckHend, line 3: we're running postgres 15.2 now

nhudson (approved):

sjmiller609 (approved):

postgres/Cargo.toml
sjmiller609: ah I think it requires to add patch version or just add publish = false

PR #199, opened 2023-04-02
sqlx-migrations

ChuckHend:
- since we're using sqlx for binding query params and compile time checks, let's also use sqlx for the migrations
- sets up the connection pool outside the route. the pool gets set up once then each route can draw a connection from the pool when it needs it, rather than creating a new connection pool on each request
- added the foreign key constraint to the versions table
- `justfile` for some dev helpers


trunk/registry/src/main.rs
```diff
@@ -1,17 +1,29 @@
...
+
+    let conn = connect(&cfg.database_url)
+        .await
+        .expect("error connecting to database");
+
+    // run database migrations
+    sqlx::migrate!()
+        .run(&conn)
+        .await
+        .expect("error running migrations");
```
ianstanton, line 20: :+1:

ianstanton (approved):
Looks good. Thanks!

PR #208, opened 2023-04-05
volumes for sharedir, pkglibdir

ChuckHend:


coredb-operator/src/defaults.rs
```diff
@@ -35,13 +35,24 @@ pub fn default_port() -> i32 {
...
+
+pub fn default_sharedir_storage() -> Quantity {
+    Quantity("250M".to_string())
+}
+
+
+pub fn default_pkglibdir_storage() -> Quantity {
+    Quantity("250M".to_string())
+}
+
```
ChuckHend, line 54: `sharedir` and `libpkgdir` are currently about 50mb each.

coredb-operator/src/statefulset.rs
```diff
@@ -182,6 +201,13 @@ pub fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
                             docker_setup_env
                             docker_create_db_directories
 
+                            cp -r /tmp/pg_sharedir/* $(pg_config --sharedir)/
+                            cp -r /tmp/pg_pkglibdir/* $(pg_config --pkglibdir)/
+                            chown -R :postgres $(pg_config --sharedir)
+                            chmod -R 2775 $(pg_config --sharedir)
+                            chown -R :postgres $(pg_config --pkglibdir)
+                            chmod -R 2775 $(pg_config --pkglibdir)
```
ChuckHend, line 209: I moved this from the `install-extension.sh` script to here, so thats its next to all the other operator specific permission changes.

coredb-operator/src/statefulset.rs
```diff
@@ -207,21 +233,53 @@ pub fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
...
+                    spec: Some(PersistentVolumeClaimSpec {
+                        access_modes: Some(vec!["ReadWriteOnce".to_owned()]),
+                        resources: Some(ResourceRequirements {
+                            limits: None,
+                            requests: Some(pvc_requests_datadir),
+                        }),
+                        ..PersistentVolumeClaimSpec::default()
+                    }),
+                    status: None,
                 },
```
ChuckHend, line 251: PVC for datadir should be exactly the same as it was before the change

coredb-operator/src/statefulset.rs
```diff
@@ -207,21 +233,53 @@ pub fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
...
-                spec: Some(PersistentVolumeClaimSpec {
-                    access_modes: Some(vec!["ReadWriteOnce".to_owned()]),
-                    resources: Some(ResourceRequirements {
-                        limits: None,
-                        requests: Some(pvc_requests),
+                PersistentVolumeClaim {
+                    metadata: ObjectMeta {
+                        name: Some("sharedir".to_string()),
```
ChuckHend, line 254: PVCs for `sharedir` and `libpkgdir` were modeled off the `datadir` PVC

coredb-operator/src/statefulset.rs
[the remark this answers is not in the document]
ChuckHend: `root` is the user that owns it but the `postgres` group has ownership. My thought was that this would let `trunk` manage all the extensions, not just the ones that are installed at runtime. If thats not what we want, then we probably only need to give the `postgres` group write permissions to: `$sharedir/extension`, `$sharedir/bitcode` (if it exists), and `$pkglibdir` 

```
-rw-r--r-- 1 root     postgres  153 Apr  5 17:58 vector--0.3.1--0.3.2.sql
```

> what about deleting the left-overs, meaning files that are in the sharedir, pkglibdirs already but originated from previous copies to the PVC?

Yes good idea. I'll handle that in a follow-up PR.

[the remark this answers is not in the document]
ChuckHend: I'll address the recursive `chown` in that next PR too

coredb-operator/src/defaults.rs
```diff
@@ -35,13 +35,24 @@ pub fn default_port() -> i32 {
...
+    "quay.io/coredb/postgres:6e3c4a7".to_owned()
 }
 
 pub fn default_storage() -> Quantity {
     Quantity("8Gi".to_string())
 }
 
+
+pub fn default_sharedir_storage() -> Quantity {
+    Quantity("250M".to_string())
```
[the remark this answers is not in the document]
ChuckHend: fixed!

coredb-operator/src/apis/coredb_types.rs
```diff
@@ -22,6 +22,10 @@ pub struct CoreDBSpec {
     pub resources: ResourceRequirements,
     #[serde(default = "defaults::default_storage")]
     pub storage: Quantity,
+    #[serde(default = "defaults::default_sharedir_storage")]
+    pub sharedir_storage: Quantity,
+    #[serde(default = "defaults::default_pkglibdir_storage")]
+    pub pkglibdir_storage: Quantity,
```
ianstanton, line 28: Nit - These are typically camelCase (`postgresExporterImage` and `apiVersion` for example).

ianstanton (approved):
Looks good, just one nit

PR #210, opened 2023-04-06
cleanup stale system files

ChuckHend:
- remove any system files in the `sharedir` and `pkglibdir` PVCs that are no longer part of the image build, on a 1-day lag.
- remove the resursive chmod and chown on the PVCs
- change some crd naming convention to address https://github.com/CoreDB-io/coredb/pull/208#discussion_r1159817506


coredb-operator/tests/integration_tests.rs
```diff
@@ -100,7 +100,7 @@ mod test {
         let replicas = 1;
 
         // Timeout settings while waiting for an event
-        let timeout_seconds_start_pod = 60;
+        let timeout_seconds_start_pod = 90;
```
ChuckHend, line 103: pod startup is slowing down. I don't think its completely related to this PR and it might be the reason for our flaky tests. I'm looking into it now.

sjmiller609: There may be some issue with no concurrency, like the operator only handling one coredb at a time. I think I recall noticing that last time I incremented this value.

sjmiller609 (approved):

PR #211, opened 2023-04-06
handle pvc updates

ChuckHend:
- handle changes to all pvcs, same as we've been doing for the data dir pvc
- all image get set to the operator's default image, rather than the currently set image on the definition
- changed signatures on the stateful set functions to reduce number of `.clone()` that get called

### DIFF
--- a/coredb-operator/src/statefulset.rs
+++ b/coredb-operator/src/statefulset.rs
@@ -1,4 +1,4 @@
-use crate::{apis::coredb_types::CoreDB, Context, Error, Result};
+use crate::{apis::coredb_types::CoreDB, Context, Error, Result, defaults::default_image, defaults::default_postgres_exporter_image};
 use k8s_openapi::{
     api::{
         apps::v1::{StatefulSet, StatefulSetSpec},
@@ -12,7 +12,7 @@ use k8s_openapi::{
 };
 use kube::{
     api::{Api, DeleteParams, ObjectMeta, Patch, PatchParams, ResourceExt},
-    Client, Resource,
+    Resource,
 };
 
 use k8s_openapi::{
@@ -20,6 +20,7 @@ use k8s_openapi::{
     apimachinery::pkg::util::intstr::IntOrString,
 };
 use std::{collections::BTreeMap, sync::Arc};
+use tracing::info;
 
 const PKGLIBDIR: &str = "/usr/lib/postgresql/15/lib";
 const SHAREDIR: &str = "/usr/share/postgresql/15";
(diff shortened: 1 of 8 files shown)


coredb-operator/src/statefulset.rs
```diff
@@ -309,12 +314,35 @@ pub async fn reconcile_sts(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error>
...
+        let sts_namespace = sts.clone().metadata.namespace.unwrap();
+
+        // why do delete first, then update?
+        delete_sts_no_cascade(&sts_api, &sts_name).await;
+
+        let pvc_api: Api<PersistentVolumeClaim> = Api::namespaced(ctx.client.clone(), &sts_namespace);
+        for (pvc_name, qty) in pvcs_to_update {
+            update_pvc(&pvc_api, &sts_name, &pvc_name, qty).await;
+        }
     }
```
ChuckHend, line 346: @sjmiller609 - can you remind me why we first delete the sts, then update the pvc?

coredb-operator/src/statefulset.rs
```diff
@@ -339,21 +363,21 @@ async fn delete_sts(client: Client, sts: &StatefulSet) {
...
         .map_err(Error::KubeError);
 }
 
-async fn update_pvc(client: Client, sts: &StatefulSet, cdb: &CoreDB) {
-    let pvc_api: Api<PersistentVolumeClaim> =
-        Api::namespaced(client, &sts.clone().metadata.namespace.unwrap());
-
-    let sts_name = sts.clone().metadata.name.unwrap();
-    let pvc_name = format!("data-{sts_name}-0");
+async fn update_pvc(pvc_api: &Api<PersistentVolumeClaim>, sts_name: &str, pvc_name: &str, value: Quantity) {
```
ChuckHend, line 371: Made this function a bit more generic, so we can use it for updating each pvc individually.

coredb-operator/src/statefulset.rs
sjmiller609: deleting the sts without cascade prevents the sts from trying to manage the pvcs it owns. Then increase the pvcs to match what they will be in the new sts, then create new sts which discovers the existing pvcs and adopts them. It might be possible to update the pvcs, then delete sts, then recreate sts, but I'm not sure if that order works.

ianstanton (approved):

coredb-operator/src/statefulset.rs
nhudson: Updating PVC's in an already defined STS is immutable and wont update.  So you are forced to delete the sts first then have the operator recreate the sts.

nhudson (approved):

PR #222, opened 2023-04-11
handle create pvc events

ChuckHend:
- delete pod after new PVCs are created

### DIFF
--- a/coredb-operator/src/statefulset.rs
+++ b/coredb-operator/src/statefulset.rs
@@ -24,7 +24,7 @@ use k8s_openapi::{
     apimachinery::pkg::util::intstr::IntOrString,
 };
 use std::{collections::BTreeMap, sync::Arc};
-use tracing::{debug, error, info};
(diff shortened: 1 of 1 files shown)


coredb-operator/src/statefulset.rs
```diff
@@ -306,79 +307,133 @@ pub fn stateful_set_from_cdb(cdb: &CoreDB) -> StatefulSet {
...
         .await
-        .map_err(Error::KubeError);
+        .map_err(Error::KubeError)?;
+    thread::sleep(Duration::from_millis(3000));
+    Ok(())
+}
+
+fn pvc_full_name(pvc_name: &str, sts_name: &str) -> String {
+    format!("{pvc_name}-{sts_name}-0")
 }
```
ChuckHend, line 429: moved this logic out, so we can create and use the pvc names upstream in the pvc reconcile process

sjmiller609 (approved):

PR #232, opened 2023-04-12
fix extension_plan params

ChuckHend:
- need to pass in diff vec and actual vec into the extension_plan function. not the desired and actual vecs.


coredb-operator/src/extensions.rs
```diff
@@ -409,7 +412,7 @@ pub async fn reconcile_extensions(coredb: &CoreDB, ctx: Arc<Context>) -> Result<
     }
 
     // otherwise, need to determine the plan to apply
-    let (changed_extensions, extensions_to_install) = extension_plan(&desired_extensions, &actual_extensions);
+    let (changed_extensions, extensions_to_install) = extension_plan(&extensions_changed, &actual_extensions);
```
ChuckHend, line 415: this was the bug :( 

the previous unit-tests were set up right, but the code was not.

https://github.com/CoreDB-io/coredb/blob/be34c6ee7de8ea128c8499e8ce6b5d910ca59ec7/coredb-operator/src/extensions.rs#L489

nhudson (approved):

PR #233, opened 2023-04-12
Update images to not use Dockerhub

sjmiller609:


ChuckHend (approved):

sjmiller609:
tried to do upload multi architecture for ubuntu:22.04 because previously we were supporting both linux/amd64 and linux/arm64 for pgx builder, so it would work on Mac. Just going to set to amd64 only for now.

PR #235, opened 2023-04-12
Bump pgmq extension dependency to 0.10.1

ianstanton:
Bump pgmq extension's crate dependency to latest (`0.10.1`).

### DIFF
--- a/extensions/pgmq/Cargo.toml
+++ b/extensions/pgmq/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.2.1"
+version = "0.3.0"
 edition = "2021"
 authors = ["CoreDB.io"]
 description = "Postgres extension for PGMQ"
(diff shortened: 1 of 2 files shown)


extensions/pgmq/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.2.1"
+version = "0.3.0"
```
ChuckHend, line 3: were there new features added to the extension? I think there were new features to the crate but they have not been added to the extension, afaik.

ianstanton: No, I guess not. What's the best way to find and add those features? I see there have been 4 bumps to the crate since the extension was last updated.

ChuckHend: If there's new new features added to the extension then this should be a patch bump to `0.2.2`

ChuckHend: I'm not sure what the _best_ way is but there were two feature releases to the crate; `send_delay` in 0.9 and `set_vt` in 0.10.

ianstanton: Okay, thanks! I'll look into adding those.

ianstanton:
Moving to draft as we bring in features from more recent crates.

ianstanton:
Closing. To be continued at a later date.

PR #239, opened 2023-04-12
Add param to trunk install cmd

ChuckHend:
Specify the registry url in the trunk install command. Let's move this to an environment variable soon.


ianstanton (approved):
Default value updated in CLI here: https://github.com/CoreDB-io/coredb/pull/238/commits/2a15286d9d491ed58886153bb120d2485afe7853

PR #255, opened 2023-04-17
Bump pgx vers trunk

ianstanton:
`trunk build` fails in our example extension project due to `pgx` version:
```
Building from path .
Detected that we are building a pgx extension
Detected pgx version range =0.7.4
thread 'main' panicked at 'error occurred: Cargo manifest error: No supported version of pgx satisfies the range =0.7.4. 
Supported versions: ["0.7.2", "0.7.1"]', /home/ian/.cargo/registry/src/github.com-1ecc6299db9ec823/pg-trunk-0.2.1/src/main.rs:51:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We can either bump supported versions (this PR), or dial back the example version to `0.7.2`.

### DIFF
--- a/trunk/cli/.github/workflows/images.yaml
+++ b/trunk/cli/.github/workflows/images.yaml
@@ -18,6 +18,8 @@ jobs:
         pgx_version:
           - "0.7.1"
           - "0.7.2"
+          - "0.7.3"
+          - "0.7.4"
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


trunk/cli/.github/workflows/images.yaml
```diff
@@ -18,6 +18,8 @@ jobs:
         pgx_version:
           - "0.7.1"
           - "0.7.2"
+          - "0.7.3"
+          - "0.7.4"
```
ianstanton, line 22: @sjmiller609 I'm not sure that we want all of these here. What do you think?

ChuckHend (approved):

PR #265, opened 2023-04-24
Trunk Makefile builder

sjmiller609:
Refactored PGX builder to work in same was as C builder. In either case, the following procedure is performed:

- Run the builder file, found in this directory https://github.com/CoreDB-io/coredb/tree/trunk-c-builder/trunk/cli/src/commands/builders
- Start a temporary container from that image
- Run the install command
- Detect what files have been installed using "docker diff" command
- Package all the newly installed files into a trunk package


trunk/cli/src/commands/containers.rs
```diff
@@ -0,0 +1,83 @@
...
+                .stop_container(&id, None)
+                .await
+                .expect("error stopping container");
+            println!("Stopped {id}");
+            task.wait().await;
+        });
+    }
+}
+
+pub async fn exec_in_container(docker: Docker, container_id: &str, command: Vec<&str>) -> Result<String, anyhow::Error> {
```
sjmiller609, line 43: adding new function for running exec in container, printing output, and collecting output

trunk/cli/tests/integration_tests.rs
```diff
@@ -33,7 +36,61 @@ fn build_extension() -> Result<(), Box<dyn std::error::Error>> {
...
+    cmd.arg("build");
+    cmd.arg("--path");
+    cmd.arg(extension_path.as_os_str());
+    cmd.arg("--output-path");
+    cmd.arg(output_dir.clone());
+    cmd.arg("--version");
+    cmd.arg("1.0.3");
+    cmd.arg("--name");
+    cmd.arg("pg_tle");
+    cmd.assert().code(0);
```
sjmiller609, line 90: this test clones pg_tle, then does `trunk build --path /to/pg_tle --version 1.0.3 --name pg_tle` and checks a trunk package was created.

TODO: test trunk install from this package and also from pgx-built package

trunk/cli/src/commands/install.rs
```diff
@@ -28,7 +28,7 @@ pub struct InstallCommand {
 }
 
 #[derive(thiserror::Error, Debug)]
-pub enum PgxInstallError {
+pub enum InstallError {
```
sjmiller609, line 31: In this file we make two changes:
- handle pathed installs
- check for architecture at the level of the whole artifact instead of each file

trunk/cli/src/commands/pgx.rs
```diff
@@ -1,36 +1,21 @@
-use bollard::container::{
```
sjmiller609, line 1: In the PGX logic, we delete everything that is not reusable and move reusable parts into containers.rs. Then we use the reusable logic that is shared with C builder in this builder.

trunk/cli/src/manifest.rs
```diff
@@ -8,7 +8,7 @@ use std::path::{Path, PathBuf};
 pub enum PackagedFile {
     ControlFile {},
     SqlFile {},
-    SharedObject { architecture: Option<String> },
+    SharedObject {},
```
sjmiller609, line 11: we set architecture at the level of the whole package instead of each file

trunk/cli/src/commands/build.rs
```diff
@@ -1,5 +1,7 @@
 use super::SubCommand;
+use crate::commands::generic_build::build_generic;
```
sjmiller609, line 2: In this file, we add a new type of build - C build

trunk/cli/src/commands/install.rs
```diff
@@ -137,17 +127,15 @@ impl SubCommand for InstallCommand {
             tempfile.write_reader(gz)?;
             tempfile.rewind()?;
             let input = tempfile;
-            install(input, extension_dir, package_lib_dir, bitcode_dir, sharedir).await?;
+            install(input, package_lib_dir, sharedir).await?;
```
sjmiller609, line 130: installation directories are being simplified to just sharedir or pkglibdir instead of each file type. File type still used to determine if it is under sharedir or pkglibdir

image

trunk/cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pg-trunk"
-version = "0.3.2"
+version = "0.4.0-alpha.1"
```
sjmiller609, line 3: we have to republish packages since they are in a new format, so I don't want it as the default on cargo install for now

ianstanton (approved):
Let's go! Awesome work @sjmiller609 :tada:

ChuckHend (approved):

PR #266, opened 2023-04-24
[pgmq] remove redundant parsing

ChuckHend:
No need to serialize from json to string, then cast back to jsonb for the insert. Instead, leave it typed jsonb and insert it directly, bypassing the string parsing. Using a bind parameter for this, instead of using the `format!` macro, which would require us to cast the `pgx::jsonb` struct to string.

### DIFF
--- a/pgmq/extension/src/lib.rs
+++ b/pgmq/extension/src/lib.rs
@@ -20,14 +20,6 @@ enum PgmqExtError {
     QueueError(#[from] PgmqError),
 }
 
-type MessageRow = (
-    i64,
-    i32,
-    TimestampWithTimeZone,
-    TimestampWithTimeZone,
-    pgx::JsonB,
-);
-
 #[pg_extern]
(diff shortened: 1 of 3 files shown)


pgmq/extension/src/lib.rs
```diff
@@ -54,18 +53,19 @@ fn pgmq_create(
...
 
-fn enqueue_str(name: &str, message: &str) -> Result<String, PgmqError> {
+fn enqueue_str(name: &str) -> Result<String, PgmqError> {
     check_input(name)?;
-    // TODO: vt should be now() + delay
     Ok(format!(
         "
         INSERT INTO {TABLE_PREFIX}_{name} (vt, message)
-        VALUES (now() at time zone 'utc', '{message}'::json)
+        VALUES (now() at time zone 'utc', $1)
```
ChuckHend, line 68: this should have been `jsonb`. But a faster way is to leave the message as `pgx::JsonB` and insert with a bind parameter.

sjmiller609 (approved):

PR #302, opened 2023-05-05
[postgres image] Update  postgres image to have wal-g and config update.

nhudson:

We need to update our base postgres image to include the binary for wal-g to enable us to have backups and restore capabilities.

This PR will also modify some of the postgres parameter settings in hopes for better  performance from each db.

- fixes # [COR-562](https://linear.app/coredb/issue/COR-562/add-wal-g-to-our-postgres-container-image)


postgres/coredb-pg-slim/Dockerfile
```diff
@@ -44,26 +55,31 @@ RUN apt-get update && apt-get install -y \
...
+  RUN git clone --branch "v${PG_VECTOR_VER}" https://github.com/pgvector/pgvector.git
+  WORKDIR /git/pgvector
+  RUN set -xe; \
+        make; \
+        make install;
+  
+  WORKDIR / 
+  RUN rm -rf /git
+  
+  RUN trunk install --version ${PG_PARTMAN_VER} pg_partman && trunk install --version ${PGMQ_VER} pgmq
```
ianstanton, line 82: I think we want to drop extensions from this image as the become available in trunk registry

nhudson: I kind of figured that, but didn't want to risk an issue.  Thanks!

ianstanton: Sounds good! We can add them back if we see issues

PR #310, opened 2023-05-08
Use correct apiGroup for the coredbs resource

nhudson:

The newly created rbac permissions that was added in #301 was attempting to access the incorrect `apiGroup` of `coredbs.coredb.io`.


coredb-operator/charts/coredb-operator/templates/cluster-role.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 rules:
   - apiGroups: ["coredb.io"]
     resources: ["coredbs", "coredbs/status"]
-    verbs: ["get", "list", "watch", "patch"]
+    verbs: ["create", "delete", "get", "list", "watch", "patch", "update"]
```
shahadarsh, line 9: As discussed, `create` & `delete` is not needed here, and `update` is only needed for `coredbs/status`. Approving it but please remove it after testing.

PR #316, opened 2023-05-09
[coredb-operator] adding in new CoreDB spec object for backups

nhudson:

This change adds in a new spec to the `CoreDB` CR which will enable us to configure and have backups running.

- fixes # [COR-759](https://linear.app/coredb/issue/COR-759/add-wal-g-environment-variables-to-deployed-statefulset)


coredb-operator/tests/integration_tests.rs
```diff
@@ -510,6 +510,68 @@ mod test {
...
+                assert_eq!(
+                    destination_path_env,
+                    Some(String::from("s3://test-bucket/coredb/test-org/test-db"))
+                );
+                assert_eq!(walg_s3_sse_env, Some(String::from("AES256")));
+            } else {
+                println!("No environment variables found in the StatefulSet's container");
+            }
+        } else {
+            println!("No container found in the StatefulSet's template spec");
```
ChuckHend, line 573: should these `panic!()`? I dont think this test will fail if those two assertions never run.

nhudson: Yeah you are correct here, I will get that updated.

ChuckHend (approved):

PR #328, opened 2023-05-12
[coredb-operator] add env config option, add initial backup after provisioning

nhudson:

The goal of the PR is to make a way for us to take an initial backup of the instances when it's created.  There are a few things going on here though.

- Created a way of defining default environment variables.  This copies pretty much what we are doing for [cp-webserver](https://github.com/CoreDB-io/control-plane/blob/main/cp-webserver/src/config.rs).  This is to make it easier to test in CI and locally by disabling the initial backup.
- Adds the code to run basically a `kubectl exec` into the postgresql pod and run a full backup using `wal-g`
- Updates the kube crate to `0.82.2` which changes a bit of how the controller functions so I updated the code to be in line with those changes.  Added a new method to shutdown more gracefully.

- fixes # [COR-780](https://linear.app/coredb/issue/COR-780/kick-off-full-backup-when-new-instance-has-started)

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -101,7 +101,7 @@ jobs:
           cargo run --bin crdgen | kubectl apply -f -
           kubectl get crds
           # Start the operator in the background
-          cargo run &
+          ENABLE_INITIAL_BACKUP=false  cargo run &
           # Run the tests
           cargo test -- --ignored --nocapture
       - name: Debugging information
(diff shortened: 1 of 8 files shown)


nhudson:
@ChuckHend per your suggestion I would like to look into adding unit tests, but without mocking the Kubernetes API, I don't know if that will be possible.  I am looking for suggestions how to achieve this.

ChuckHend (approved):

ChuckHend (approved):

PR #333, opened 2023-05-16
[postgres-image] Ensure that the latest postgresql.conf file is copied

nhudson:
Currently the `postgresql.conf` file isn't being updated or copied over after the initial setup.  If any new config options are applied, a newer Docker image will not copy them over.

Added a new function that will always check and copy over the `postgresql.conf.sample` into the `$PGDATA/postgresql.conf` .

Adding in wal recovery options as this will be needed if we have to manually restore for now.

fixes: [COR-815](https://linear.app/coredb/issue/COR-815)
related: [COR-566](https://linear.app/coredb/issue/COR-566)

### DIFF
--- a/postgres/coredb-pg-slim/Cargo.toml
+++ b/postgres/coredb-pg-slim/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "coredb-pg-slim"
-version = "15.2.0-coredb-pg-slim.9"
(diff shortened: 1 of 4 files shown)


ianstanton (approved):

nhudson:
Will have to make additional changes.  We can't source the configuration from `/usr/share/postgresql/15/postgresql.conf.sample` since that directory is also mounted as a persistent volume.

PR #335, opened 2023-05-16
fix function to update config

nhudson:
While what was added earlier works for images where the `$PGDATA` directory is already in place, it did not work for a new instance.  This should fix that


nhudson:
Hopefully this will fix https://github.com/CoreDB-io/coredb/pull/334

ChuckHend (approved):

PR #343, opened 2023-05-19
configmap for custom pg-prom metric queries

ChuckHend:


coredb-operator/charts/coredb-operator/templates/crd.yaml
```diff
@@ -93,6 +93,10 @@ spec:
                   image:
                     default: quay.io/prometheuscommunity/postgres-exporter:v0.12.0
                     type: string
+                  queries:
+                    x-kubernetes-preserve-unknown-fields: true
```
nhudson, line 97: Why do we need this?  I don't think I understand.

ChuckHend: We don't know the metric config names that come in queries.yaml, we don't know the keys ahead of time. e.g. the `pg_postmaster` in the yaml below - thats the "key", and that can be arbitrarily named. We don't know all the names it will be, so we can't set a struct attribute to "pg_postmaster". So in this case `pg_postmaster` is the unknown-field that we want to preserve.

```yaml
pg_postmaster:
  query: SELECT pg_postmaster_start_time as start_time_seconds from pg_postmaster_start_time()
  master: true
  metrics:
  - start_time_seconds:
      usage: Gauge
      description: Time at which postmaster started
 ```

Comment (ChuckHend) [reply coredb-operator/charts/coredb-operator/templates/crd.yaml:97]: without it we get
```
strict decoding error: unknown field "spec.metrics.queries. pg_postmaster"
```

Comment (nhudson) [reply coredb-operator/charts/coredb-operator/templates/crd.yaml:97]: Ah ok. Thanks for the explanation, that is correct then.

File: coredb-operator/src/controller.rs [hunk 2]
```diff
@@ -209,12 +219,7 @@ impl CoreDB {
                 storage: self.spec.storage.clone(),
                 sharedirStorage: self.spec.sharedirStorage.clone(),
                 pkglibdirStorage: self.spec.pkglibdirStorage.clone(),
-                extensions: self
-                    .status
-                    .as_ref()
-                    .expect("no extensions in `status`")
-                    .extensions
-                    .clone(),
+                extensions: self.status.clone().and_then(|f| f.extensions),
```
ChuckHend, line 222: this was a bug. this shouldn't panic when there are no extensions in status. its an `Option`, and we should allow it to be `Option::None`

sjmiller609 (approved):

PR #344, opened 2023-05-23
Postgres Base Image

ianstanton:
Create a base image with Postgres and no extras:
- Postgres 15.3 compiled and installed from source
- No extensions included (other than `plpgsql`)
    ```
    postgres@84bb4052d43b:/postgresql-15.3$ psql
    psql (15.3)
    Type "help" for help.
    
    postgres=# select * from pg_available_extensions;
      name   | default_version | installed_version |           comment            
    ---------+-----------------+-------------------+------------------------------
     plpgsql | 1.0             | 1.0               | PL/pgSQL procedural language
    ```
- No trunk, walg, etc.


nhudson (commented):
My only suggestion is to add [this](https://github.com/CoreDB-io/coredb/blob/af5f64dc4c45dcb4104d34acc15629e6bd8f7837/postgres/coredb-pg-slim/docker-entrypoint.sh#L251) to the `docker-entrypoint.sh`.  

This will ensure that any changes in the `postgresql.conf` file is updated if it's updated in this base image.

nhudson (changes requested):
That function will need to be called [here](https://github.com/CoreDB-io/coredb/blob/af5f64dc4c45dcb4104d34acc15629e6bd8f7837/postgres/coredb-pg-slim/docker-entrypoint.sh#L342-L343) as well

nhudson (approved):

PR #356, opened 2023-06-06
relative dependency in extension build

ChuckHend:


pgmq/extension/Cargo.toml
```diff
@@ -25,7 +25,7 @@ pg_test = []
 [dependencies]
 pgrx = "0.7.4"
 serde = "1.0.152"
-pgmq_crate = {package = "pgmq", version = "0.11.4"}
+pgmq_crate = {package = "pgmq", path = "../core"}
```
ChuckHend, line 28: this is the reason for the change to how we execute `trunk build`

PR #373, opened 2023-06-26
add modified hba conf file to allow for migration to cnpg

nhudson:
Copy in modified `pg_hba.conf` file into the now old CoreDB image.  This will allow us to bootstrap CNPG instances from the old CoreDB instances using automation inside CNPG.

fixes: [TEM-1192](https://linear.app/tembo/issue/TEM-1192/hba-conf-in-coredb-for-replication)


sjmiller609 (approved):
looks good, this hba conf is temporary for the cut over.

PR #396, opened 2023-12-05
Conditionally apply IngressRouteTCP

EvanHStanton:
### Routing Configuration Enhancements (types.rs):
- Added entryPoints property to the Routing struct, enabling the specification of entry points for routing.

### Ingress Route TCP Implementation (ingress.rs):
- Implemented generate_ingress_tcp and convert_routes_to_tcp functions to facilitate TCP routing.
- Enhanced reconcile_ingress function to support both IngressRoute and IngressRouteTCP based on the presence of specific entry points.
- Added apply_ingress_route_tcp asynchronous function for applying TCP ingress routes.

### App Service Resource Management (manager.rs):
- Modified AppServiceResources struct to include entry_points.
- Updated generate_resource function to fetch entry points from routing configuration.
- Adjusted reconcile_app_services function to incorporate desired entry points in the ingress reconciliation process.

### DIFF
--- a/tembo-operator/src/app_service/ingress.rs
+++ b/tembo-operator/src/app_service/ingress.rs
@@ -24,6 +24,11 @@ use super::{
     types::{AppService, Middleware, COMPONENT_NAME},
 };
 
+use crate::traefik::ingress_route_tcp_crd::{
+    IngressRouteTCP, IngressRouteTCPRoutes, IngressRouteTCPRoutesMiddlewares, IngressRouteTCPRoutesServices,
+    IngressRouteTCPSpec, IngressRouteTCPTls,
+};
+
 #[derive(Clone, Debug)]
 pub struct MiddleWareWrapper {
     pub name: String,
@@ -57,6 +62,41 @@ fn generate_ingress(
             entry_points: Some(vec!["websecure".to_string()]),
             routes,
             tls: Some(IngressRouteTls::default()),
+        },
+    }
+}
+
+fn generate_ingress_tcp(
+    coredb_name: &str,
+    namespace: &str,
+    oref: OwnerReference,
+    routes: Vec<IngressRouteTCPRoutes>,
+    entry_points: Vec<String>,
+) -> IngressRouteTCP {
+    let mut selector_labels: BTreeMap<String, String> = BTreeMap::new();
+
+    selector_labels.insert("component".to_owned(), COMPONENT_NAME.to_string());
+    selector_labels.insert("coredb.io/name".to_owned(), coredb_name.to_string());
+
+    let mut labels = selector_labels.clone();
+    labels.insert("component".to_owned(), COMPONENT_NAME.to_owned());
+
+    IngressRouteTCP {
+        metadata: ObjectMeta {
+            // using coredb name, since we'll have 1x ingress per coredb
+            name: Some(coredb_name.to_owned()),
+            namespace: Some(namespace.to_owned()),
+            owner_references: Some(vec![oref]),
+            labels: Some(labels.clone()),
+            ..ObjectMeta::default()
+        },
+        spec: IngressRouteTCPSpec {
+            entry_points: Some(entry_points),
+            routes,
+            tls: Some(IngressRouteTCPTls {
+                passthrough: Some(true),
+                ..IngressRouteTCPTls::default()
+            }),
         },
     }
 }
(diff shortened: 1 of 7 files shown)


tembo-operator/src/app_service/ingress.rs
```diff
@@ -61,6 +66,41 @@ fn generate_ingress(
     }
 }
 
+fn generate_ingress_tcp(
```
EvanHStanton, line 69: New function for generating `IngressRouteTCP`.

tembo-operator/src/app_service/ingress.rs
```diff
@@ -168,6 +200,11 @@ pub fn generate_ingress_routes(
             for route in routings.iter() {
                 match route.ingress_path.clone() {
                     Some(path) => {
+                        if route.entry_points.clone()?.contains(&"ferretdb".to_string()) {
```
EvanHStanton, line 203: If the entry point contains "ferretdb" then the ingress route will not be generated.

tembo-operator/src/ingress.rs
```diff
@@ -270,8 +271,24 @@ pub async fn reconcile_postgres_ing_route_tcp(
     // Check for all existing IngressRouteTCPs in this namespace
     // Filter out any that are not for this DB or do not have the correct prefix
     for ingress_route_tcp in &ingress_route_tcps {
```
ianstanton, line 273: There was some conflict between ingress reconciliation in this file and in app services, so we added some logic for ignoring ingress with label `component: appService`

tembo-operator/src/app_service/types.rs
```diff
@@ -76,6 +76,28 @@ pub struct Routing {
...
+    #[serde(default = "default_ingress_type")]
+    pub ingress_type: Option<IngressType>,
+}
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize, ToSchema, JsonSchema)]
+pub enum IngressType {
+    http,
+    tcp,
+    udp,
```
ianstanton, line 92: `udp` doesn't do anything for now. We could drop this and add in the future if we like.

tembo-operator/tests/integration_tests.rs
```diff
@@ -3029,8 +3067,8 @@ mod test {
         // Fetch and print all the pods matching the label selector
         let pod_list = pods.list(&lp).await.unwrap();
         assert_eq!(pod_list.items.len(), 1);
-        let app_0_pod = pod_list.items[0].clone();
-        let app_0_container = app_0_pod.spec.unwrap().containers[0].clone();
+        let app_1_pod = pod_list.items[0].clone();
+        let app_1_container = app_1_pod.spec.unwrap().containers[0].clone();
```
ianstanton, line 3071: Order of the apps changed after introducing a new app (ferretdb)

tembo-operator/src/app_service/types.rs
ChuckHend: lets drop it if it doesnt do anything

ChuckHend, line 92:
```suggestion
```

tembo-operator/tests/integration_tests.rs
```diff
@@ -2958,7 +2958,11 @@ mod test {
                         "routing": [
                             {
                                 "port": 3000,
-                                "ingressPath": "/"
+                                "ingressPath": "/",
+                                "entryPoints": [
+                                    "websecure"
+                                ],
```
ChuckHend, line 2964: existing app_services do not need to start including a value for `entryPoints` right? Let's leave the test using the default since there's already services that exist that assume it will default to `websecure`

tembo-operator/src/app_service/types.rs
ianstanton: :+1: @EvanHStanton took care of this

tembo-operator/tests/integration_tests.rs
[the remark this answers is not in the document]
EvanHStanton: https://github.com/tembo-io/tembo/pull/396/commits/e24d71eee8cdcb04c611d51e6c1bbb0d3956e6bf

ChuckHend (approved):

tembo-operator/src/app_service/ingress.rs
```diff
@@ -54,13 +58,44 @@ fn generate_ingress(
...
+    entry_points: Vec<String>,
+) -> IngressRouteTCP {
+    let mut labels: BTreeMap<String, String> = BTreeMap::new();
+    labels.insert("component".to_owned(), COMPONENT_NAME.to_string());
+    labels.insert("coredb.io/name".to_owned(), coredb_name.to_string());
+
+    IngressRouteTCP {
+        metadata: ObjectMeta {
+            // using coredb name, since we'll have 1x ingress per coredb
+            name: Some(coredb_name.to_owned()),
```
sjmiller609, line 82: can you add a short suffix like "-apps" because for old servers, this name is already used for the ingressroute tcp that ingress postgres

ianstanton: I was thinking that as well and added the label logic to ignore them in the postgres ingress reconciliation. I'm up for adding this as well though.

ianstanton: I had tried this but ran into an issue so punted on it. @EvanHStanton can you try this out and see what happens?

ianstanton: Updated

sjmiller609 (approved):

ianstanton (approved):

PR #400, opened 2023-12-06
Custom metrics

sjmiller609:
Setting up a way to quickly add additional SQL-based metrics

### DIFF
--- a/tembo-operator/src/configmap.rs
+++ b/tembo-operator/src/configmap.rs
@@ -1,12 +1,70 @@
-use crate::Error;
+use crate::{apis::coredb_types::CoreDB, Context, Error};
 use k8s_openapi::api::core::v1::ConfigMap;
 use kube::{
     api::{Api, ObjectMeta, Patch, PatchParams},
-    Client,
+    runtime::controller::Action,
+    Client, ResourceExt,
 };
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, sync::Arc};
 
-use tracing::{debug, error};
+use tracing::{debug, error, instrument};
+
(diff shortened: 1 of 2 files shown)


tembo-operator/src/configmap.rs
```diff
@@ -1,12 +1,70 @@
...
+            return None;
+        }
+    };
+    let custom_metrics_name = match std::env::var("CUSTOM_METRICS_CONFIGMAP_NAME") {
+        Ok(name) => name,
+        Err(_) => {
+            debug!("CUSTOM_METRICS_NAME not set, skipping adding custom metrics");
+            return None;
+        }
+    };
```
sjmiller609, line 65:
```suggestion
    let custom_metrics_namespace = match std::env::var("CUSTOM_METRICS_CONFIGMAP_NAMESPACE") {
        Ok(namespace) => namespace,
        Err(_) => {
            debug!("CUSTOM_METRICS_CONFIGMAP_NAMESPACE not set, skipping adding custom metrics");
            return None;
        }
    };
    let custom_metrics_name = match std::env::var("CUSTOM_METRICS_CONFIGMAP_NAME") {
        Ok(name) => name,
        Err(_) => {
            debug!("CUSTOM_METRICS_CONFIGMAP_NAME not set, skipping adding custom metrics");
            return None;
        }
    };
```

nhudson (approved):

ChuckHend (approved):
Where would one define the custom metrics?

sjmiller609:
> Where would one define the custom metrics?

This allows a user of tembo operator to specify a configmap to pull metrics from. In the case of Tembo, it's here https://github.com/tembo-io/charts/blob/main/alerts/files/postgres-metrics.yaml

PR #402, opened 2023-12-07
adds tembo.toml example & fixes clippy error

shahadarsh:


tembo-cli/examples/context
```diff
@@ -0,0 +1,12 @@
+version = "1.0"
+
+[[environment]]
```
shahadarsh, line 3: This will be changed to use like below soon.

```
[local]
name = "local"
target = "docker"
[prod]
...
```

PR #411, opened 2023-12-10
adds support for updating & deleting instance on tembo-cloud

shahadarsh:


tembo-cli/src/cli/context.rs
```diff
@@ -88,6 +88,14 @@ pub fn tembo_credentials_file_path() -> String {
     tembo_home_dir() + "/credentials"
 }
 
+pub fn dot_tembo_folder() -> String {
+    ".tembo".to_string()
```
DarrenBaldwin07, line 92: Cant these just be a `const DOT_TEMBO_FOLDER: String = String::from(".tembo.io")`?

DarrenBaldwin07 (approved):

tembo-cli/src/cli/context.rs
shahadarsh: Yup, good catch. Changed it.

PR #414, opened 2023-12-11
Expose storageClass in `CoreDB`

nhudson:
Allow configuration of `storageClass` in the `CoreDB` API.  This will allow other users to use whatever storage class they wish inside their cluster.  Currently we only support the `default` storage class.

* Fix clippy warnings in the code as well.

Fixes: [TEM-2582](https://linear.app/tembo/issue/TEM-2582/support-storageclass-in-coredb)


nhudson:
Once this is merged we can add a note and explanation of how to set this in the documentation https://github.com/tembo-io/website/pull/221

sjmiller609 (approved):

PR #416, opened 2023-12-12
Ensure that the initial backups are completed before reconciliation finish

nhudson:
If you have a scenario where you are restoring a large database the backup may take a very long time.  We need to hold off on full reconciliation until the backup is complete.

This adds some functionality to check  for backups upon the first reconciliation loop to ensure that the initial backup is ran and completes successfully before we finish a reconcile.

Fixes: [TEM-2594](https://linear.app/tembo/issue/TEM-2594/large-restores-restart-in-a-loop)

Todo: 
* Writing/Updating tests.

### DIFF
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -1736,6 +1736,11 @@ pub async fn check_backups_status(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(),
 
     match backup_result {
         Ok(backup_list) => {
+            if backup_list.items.is_empty() {
+                error!("No backups found for {}, requeuing", instance_name);
+                return Err(Action::requeue(Duration::from_secs(30)));
+            }
+
             for backup_item in backup_list.items {
                 if let Some(backup_name) = &backup_item.metadata.name {
                     if !is_backup_completed(&backup_item) {
@@ -1743,8 +1748,6 @@ pub async fn check_backups_status(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(),
                         return Err(Action::requeue(Duration::from_secs(30)));
                     }
                     info!("Backup {} completed", backup_name);
-                } else {
-                    warn!("Found backup with no name");
                 }
             }
             Ok(())
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -4,6 +4,7 @@ use crate::{
         postgres_parameters::MergeError,
     },
     cloudnativepg::{
+        backups::Backup,
         clusters::{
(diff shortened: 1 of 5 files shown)


tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -729,9 +730,8 @@ async fn pods_to_fence(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<String>, A
     if cdb.spec.restore.is_some()
         && cdb
             .status
-            .clone()
-            .unwrap_or_default()
-            .first_recoverability_time
+            .as_ref()
+            .and_then(|s| s.first_recoverability_time.as_ref())
```
nhudson, line 734: This is a less expensive way of looking up `first_recoverability_time`

tembo-operator/src/controller.rs
```diff
@@ -309,6 +312,22 @@ impl CoreDB {
...
+                        .and_then(|s| s.first_recoverability_time)
+                        .is_none()
+                {
+                    check_backups_status(self, ctx.clone()).await?;
+                    // if let Err(e) = check_backups_status(self, ctx.clone()).await {
+                    //     error!("Error checking backups status: {:?}", e);
+                    //     return Err(Action::requeue(Duration::from_secs(300)));
+                    // }
+                }
+
```
nhudson, line 330: Not sure if this is the best place, but ideally it needs to be the last thing ran before we wrap up reconciliation.  We also might want to do this on restores only and not initial instances.

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1713,6 +1713,49 @@ fn generate_s3_restore_credentials(
...
+// complete, we will requeue until it is.
+pub async fn check_backups_status(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Action> {
+    let instance_name = cdb.name_any();
+    let namespace = cdb.namespace().ok_or_else(|| {
+        error!("Namespace is not set for CoreDB for instance {}", instance_name);
+        Action::requeue(Duration::from_secs(300))
+    })?;
+
+    let backups_api: Api<Backup> = Api::namespaced(ctx.client.clone(), &namespace);
+    let label_selector = format!("cnpg.io/cluster={},cnpg.io/immediateBackup=true", instance_name);
```
sjmiller609, line 1733: I think after 30 days, existing instances will not have any backups with immediateBackup=true, maybe we just check for any backup completed, corresponding to this instance, and completed in the last 48 hours for this step.
```
  labels:
    cnpg.io/cluster: org-***-inst-***
    cnpg.io/immediateBackup: "false"
    cnpg.io/scheduled-backup: org-***-inst-***
    ```

sjmiller609 (approved):

PR #417, opened 2023-12-12
Mount certs from secret in app services

ianstanton:
Some app services need to mount values from a secret `<coredb-name>-server1`. We don't have a good way to template the secret name into our Stacks definition, so adding this as a volume / volume mount for all app services. In the future this could be conditional.


sjmiller609 (approved):
Note, this will replace all app service pods in cloud, which should be fine but worth checking on

EvanHStanton (approved):

nhudson (approved):

PR #419, opened 2023-12-14
update how we handle fencing restore instances

nhudson:
There is still an issue where a restore instance will get fenced even after it's fully reconciled.  This is a new attempt to fix the issue.

The issue complicated to explain without showing what happens in real time, but I will do my best

- Restore initiated and the operator starts the reconciliation process

- The operator fences the pod because it's a restore.  This is done right after we call `reconcile_cnpg`.  This logic currently looks for restore is present and `first_recoverability_time` is empty, which at this time is true and should be true.

- Once the restore is complete, the backup starts up but if it's a large DB this could take some time, as you know.
 
- Due to the backup time `first_recoverability_time` will still be empty on subsequent reconcile loops.  Though currently there is [this logic](https://github.com/tembo-io/tembo/blob/a377faa100b7b17cb8094680e0f551d9feb04b9f/tembo-operator/src/controller.rs#L317-L326) I put in place with #416 , it is removed in this PR.
 
- This appears to work (at least checking for the backup being complete), but for some reason it completes and the `let recovery_time = self.get_recovery_time(ctx.clone()).await?;` call still returns a `None` value for `first_recoverability_time`.  Maybe this happens too fast?  It's happening every time, regardless.
 
- Upon the next reconciliation the pod gets fenced again due to `first_recoverability_time` not being set.  Once reconciled again, then the issue is fixed.  This is why when restoring a large instance, the instance will restart again after the backup is completed.

### DIFF
--- a/tembo-operator/Cargo.lock
+++ b/tembo-operator/Cargo.lock
@@ -494,7 +494,7 @@ dependencies = [
 
 [[package]]
 name = "controller"
-version = "0.26.1"
+version = "0.27.0"
 dependencies = [
  "actix-web",
  "anyhow",
--- a/tembo-operator/Cargo.toml
+++ b/tembo-operator/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.26.1"
+version = "0.27.0"
 edition = "2021"
 default-run = "controller"
 license = "Apache-2.0"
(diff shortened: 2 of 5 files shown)


sjmiller609 (approved):
I think this seems sufficiently safe, but it is maybe more complicated than it needs to be in the conditional fencing. Maybe we can revisit the logic of when a pod is fenced, perhaps there is a simpler set of conditions. For example if the server younger than the previous full reconciliation, and this server does not include a requested trunk install.

PR #420, opened 2023-12-14
adds support for stacks for target docker

shahadarsh:
* @sjmiller609 I used the controller crate for now. Tried using cargo workspace but it wasn't straightforward. I think we should move to use cargo workspace after we move stacks out of the operator code base. 
* Looked at merge logic in control plane. For cli I wrote something simple for `trunk_installs`, `extensions` & `postgres_config` since they need to be handled differently. See template [here](. .https://github.com/tembo-io/tembo/pull/420/files#diff-742edd85af3e0e60feb48d4c13671c80b951b2639bc44fcad587eeb65be33b46L3).

TODO:
* Handle duplicate values for `trunk_installs`, `extensions` & `postgres_config` in stack & instance.

### DIFF
--- a/tembo-cli/Cargo.lock
+++ b/tembo-cli/Cargo.lock
@@ -318,10 +318,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-trait"
+version = "0.1.74"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


sjmiller609 (approved):
I think reasonable to get working in naive case then follow up on QA and polish

shahadarsh:
> I think reasonable to get working in naive case then follow up on QA and polish

Yup that's the idea.

vrmiguel (approved):

PR #425, opened 2023-12-15
Consolidate stack metrics into metrics sidecar instead of postgres exporter deployment

nhudson:
This PR removes the `Deployment`, `Service`, `ServiceAccount`, `Role` and `RoleBinding` of the Postgres exporter from the controller.  

- For now I have added new functions to remove the above objects from the cluster if they exist.
- I have commented out the code that generated these objects for now, we can remove them all in a new PR, after this PR is merged and rolled out.
- All metrics now get mapped into the exporter that is already running in the `postgres` container setup by CNPG.

Fixes: [TEM-2542](https://linear.app/tembo/issue/TEM-2542/consolidate-stack-metrics-into-metrics-sidecar-instead-of-postgres)

### DIFF
--- a/charts/tembo-operator/Chart.yaml
+++ b/charts/tembo-operator/Chart.yaml
@@ -3,7 +3,7 @@ name: tembo-operator
 description: 'Helm chart to deploy the tembo-operator'
 type: application
 icon: https://cloud.tembo.io/images/TemboElephant.png
-version: 0.2.2
+version: 0.2.3
 home: https://tembo.io
 sources:
   - https://github.com/tembo-io/tembo-stacks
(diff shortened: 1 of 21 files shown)


nhudson:
There are a few outstanding issues with this PR that needs to be addressed

1: The integration tests will fail until I can go through all the tests that check anything to do with the old exporter and remove or cleanup the code.
2: There is a permissions issue that we need to work out.  That is laid out here https://linear.app/tembo/issue/TEM-2542/consolidate-stack-metrics-into-metrics-sidecar-instead-of-postgres#comment-7d6bb113

sjmiller609 (approved):

PR #431, opened 2023-12-18
Update matcher

EvanHStanton:
### Context:
- While developing our IngressRouteTCP we have found a need to update the matcher to adjust for non-HTTP traffic management, in this case HostSNI (Server Name Indication). This is related to our work with FerretDB, but also adds flexibility and security to future projects.

### Update IngressRouteTCP Matcher (ingress.rs)
- Update fn `generate_ingress_tcp_routes` matcher to HostSNI

### Update naming convention to be
- Update fn `reconcile_app_services` to define app_name

### Update assertion (integration_tests.rs)


tembo-operator/src/app_service/ingress.rs
```diff
@@ -269,7 +269,7 @@ pub fn generate_ingress_tcp_routes(
                             continue;
                         }
 
-                        let matcher = format!("{host_matcher} && PathPrefix(`{}`)", path);
+                        let matcher = format!("HostSNI(`{}`)", host_matcher);
```
ianstanton, line 272: You may need to update some assertions in the `functional_test_app_service` integration test

tembo-operator/src/app_service/ingress.rs
```diff
@@ -398,9 +398,10 @@ pub async fn reconcile_ingress_tcp(
     desired_routes: Vec<IngressRouteTCPRoutes>,
     desired_middlewares: Vec<Middleware>,
     entry_points_tcp: Vec<String>,
+    app_name: &str,
 ) -> Result<(), kube::Error> {
     let ingress_api: Api<IngressRouteTCP> = Api::namespaced(client.clone(), ns);
-    let name = format!("{}-apps", coredb_name);
+    let name = format!("{}-{}", coredb_name, app_name);
```
ianstanton, line 404: :ok_hand:

tembo-operator/src/app_service/manager.rs
```diff
@@ -692,29 +702,33 @@ pub async fn reconcile_app_services(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(
...
-        Ok(_) => {
-            debug!("Updated/applied IngressRouteTCP for {}.{}", ns, coredb_name,);
-        }
-        Err(e) => {
-            error!(
-                "Failed to update/apply IngressRouteTCP {}.{}: {}",
-                ns, coredb_name, e
-            );
-            has_errors = true;
+    for appsvc in appsvcs.iter() {
```
ianstanton, line 705: As a follow-up, let's make sure the expected `IngressRouteTCP` resources are created when we have multiple apps that use `IngressType: tcp`

tembo-operator/src/app_service/ingress.rs
```diff
@@ -269,7 +269,7 @@ pub fn generate_ingress_tcp_routes(
                             continue;
                         }
 
-                        let matcher = format!("{host_matcher} && PathPrefix(`{}`)", path);
+                        let matcher = format!("{}", host_matcher_tcp);
```
ChuckHend, line 272: Can we just pass `host_matcher_tcp` directly to the IngressRouteTCPRoutes, instead of re-assigning it here. There's no changes being made to it, right?

Something similar to this...

```rust
                        let route = IngressRouteTCPRoutes {
                            r#match: host_matcher_tcp,
                            services: Some(vec![IngressRouteTCPRoutesServices {
                                name: resource_name.to_string(),
```

ianstanton (approved):

ChuckHend (approved):

PR #441, opened 2023-12-20
Tembo CLI help messages

sjmiller609:
For example, tembo context --help will now show
```
Manage Tembo contexts

Usage: tembo context <COMMAND>

Commands:
  list  List all available contexts
  set   Set the current context
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```


tembo-cli/src/cli/context.rs
```diff
@@ -88,7 +87,7 @@ pub fn tembo_credentials_file_path() -> String {
     tembo_home_dir() + "/credentials"
 }
 
-pub fn list_context() -> Result<Context> {
+pub fn list_context() -> Result<Context, anyhow::Error> {
```
sjmiller609, line 90: `Result<T, E>` Type: In Rust, `Result` is an enum used for error handling. It has two variants: `Ok(T)`, which represents success and contains a value of type T, and `Err(E)`, which represents an error and contains an error of type `E`.

`anyhow::Error`: anyhow is a crate in Rust that provides a convenient way to handle errors. `anyhow::Error` is commonly used when you don't need or want to create specific error types for each kind of error.

tembo-cli/src/main.rs
```diff
@@ -1,112 +1,60 @@
...
-        Some(("delete", sub_matches)) => cmd::delete::execute(sub_matches),
-        Some(("completions", sub_matches)) => (|| {
-            let shell = sub_matches
-                .get_one::<Shell>("shell")
-                .with_context(|| "Shell name missing.")?;
+#[derive(Parser)]
+#[clap(author = crate_authors!("\n"), version = crate_version!(), about = "Tembo CLI", long_about = None)]
+struct App {
+    #[clap(flatten)]
+    global_opts: GlobalOpts,
```
sjmiller609, line 13: There were some issues with input validation of commands in subcommands, for example showing help if you have a non-existing subcommand of a subcommand. So I've tidied this code to match the subcommands example in the docs here

https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_2/index.html#subcommands

tembo-cli/src/cmd/context.rs
```diff
@@ -1,27 +1,22 @@
...
 pub mod set;
 
-// handles all context command calls
-pub fn execute(args: &ArgMatches) -> Result<()> {
-    // execute the context subcommands
-    let res = match args.subcommand() {
-        Some(("list", sub_matches)) => list::execute(sub_matches),
-        Some(("set", sub_matches)) => set::execute(sub_matches),
-
-        _ => unreachable!(),
```
sjmiller609, line 16: this is an example why refactoring to use input validation of subcommands: the command "tembo context asdf" would just have panic on unreachable code. Instead, now it will show help like this:
```
error: unrecognized subcommand 'asdfasdf'

Usage: tembo context <COMMAND>

For more information, try '--help'.
```

tembo-cli/src/main.rs
sjmiller609: For example, `tembo context --help` will now show:

```
Manage Tembo contexts

Usage: tembo context <COMMAND>

Commands:
  list  List all available contexts
  set   Set the current context
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

tembo-cli/src/cli/context.rs
shahadarsh: Nice! Thanks for the explanation.

tembo-cli/src/main.rs
shahadarsh: Awesome! Thank you.

shahadarsh (approved):
Thank you for making these changes. Looks great. :thumbsup:

DarrenBaldwin07 (approved):
Much better!

PR #443, opened 2023-12-21
change ingress path on vector db app

ChuckHend:
ergonomics are weird. the public facing endpoint ends up being `{domain}/embeddings/v1/embeddings`. This will make it `/vectordb/v1/embeddings` instead.

### DIFF
--- a/tembo-operator/src/stacks/templates/vectordb.yaml
+++ b/tembo-operator/src/stacks/templates/vectordb.yaml
@@ -7,14 +7,14 @@ appServices:
     name: embeddings
     routing:
       - port: 3000
-        ingressPath: /transform
+        ingressPath: /vectordb
         middlewares:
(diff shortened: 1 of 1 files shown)


tembo-operator/src/stacks/templates/vectordb.yaml
```diff
@@ -7,14 +7,14 @@ appServices:
...
-        ingressPath: /embeddings
+        ingressPath: /vectordb
         middlewares:
           - map-embeddings
     middlewares:
       - !replacePathRegex
           name: map-embeddings
           config:
-            regex: \/embeddings\/?
+            regex: ^\/vectordb\/?
```
ChuckHend, line 17: only match on first occurrence

ianstanton (approved):

PR #448, opened 2023-12-21
runs sqlx migration for tembo cloud deployment

shahadarsh:


tembo-cli/src/cli/sqlx_utils.rs
```diff
@@ -0,0 +1,48 @@
+use crate::Result;
+use base64::{engine::general_purpose, Engine as _};
+use spinners::Spinner;
+use spinners::Spinners;
+use sqlx::migrate::Migrator;
```
sjmiller609, line 5: 👀

tembo-cli/src/cli/sqlx_utils.rs
```diff
@@ -0,0 +1,48 @@
...
+        m.run(&pool).await?;
+
+        sp.stop_with_message("- SQL migration completed".to_string());
+
+        Ok(())
+    }
+
+    fn b64_decode(b64_encoded: &str) -> String {
+        let bytes = general_purpose::STANDARD.decode(b64_encoded).unwrap();
+        std::str::from_utf8(&bytes).unwrap().to_owned()
```
vrmiguel, line 46: In order to reuse the `Vec` we can do something like:

```suggestion
        String::from_utf8(bytes).unwrap()
```

tembo-cli/src/cli/sqlx_utils.rs
```diff
@@ -0,0 +1,48 @@
+use crate::Result;
+use base64::{engine::general_purpose, Engine as _};
+use spinners::Spinner;
+use spinners::Spinners;
+use sqlx::migrate::Migrator;
+use sqlx::Pool;
+use sqlx::Postgres;
+use std::path::Path;
+use temboclient::models::ConnectionInfo;
```
sjmiller609, line 9: Shoot - small issue. We want to use the data plane API instead of the control plane API for the username / password info because we are planning to remove from control plane the database passwords so that we don't need to save them. I.e. we are planning for control plane API to not return password / username information.

sjmiller609 (approved):

tembo-cli/src/cli/sqlx_utils.rs
shahadarsh: Ah yes, thanks for the reminder. You had mentioned that when I was implementing the TF provider. I will change it.

tembo-cli/src/cmd/apply.rs
```diff
@@ -142,6 +167,31 @@ pub fn get_instance_id(
     Ok(None)
 }
 
+pub fn is_instance_up(
+    instance_id: String,
+    config: &Configuration,
+    env: &Environment,
+) -> Result<Option<Box<ConnectionInfo>>> {
```
vrmiguel, line 174: has a `Result` return but never returns an error

vrmiguel (approved):

tembo-cli/src/cli/sqlx_utils.rs
sjmiller609: you'll want to use this one https://api.data-1.use1.tembo.io/swagger-ui/#

tembo-cli/src/cmd/apply.rs
shahadarsh: What would you recommend the best way for me to return the error here?

tembo-cli/src/cli/sqlx_utils.rs
shahadarsh: Do you mind if I merge this PR and replace the call in the next one?

PR #450, opened 2023-12-21
Add config to set target database for metric query

nhudson:
To fix issues where the default query for metrics is using the correct database.

This config is taken directly from the [CNPG docs](https://cloudnative-pg.io/documentation/1.20/monitoring/#example-of-a-user-defined-metric-running-on-multiple-databases)

Fixes: [TEM-2666](https://linear.app/tembo/issue/TEM-2666/add-target-database-for-metrics-query)

### DIFF
--- a/tembo-operator/src/app_service/ingress.rs
+++ b/tembo-operator/src/app_service/ingress.rs
@@ -262,7 +262,7 @@ pub fn generate_ingress_tcp_routes(
             let mut routes: Vec<IngressRouteTCPRoutes> = Vec::new();
             for route in routings.iter() {
                 match route.ingress_path.clone() {
-                    Some(path) => {
+                    Some(_path) => {
                         if !route.ingress_type.clone()?.eq(&IngressType::tcp) {
                             // Do not create IngressRouteTCPRoutes for non-TCP ingress type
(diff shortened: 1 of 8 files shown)


tembo-operator/src/app_service/ingress.rs
```diff
@@ -262,7 +262,7 @@ pub fn generate_ingress_tcp_routes(
             let mut routes: Vec<IngressRouteTCPRoutes> = Vec::new();
             for route in routings.iter() {
                 match route.ingress_path.clone() {
-                    Some(path) => {
+                    Some(_path) => {
```
nhudson, line 265: This is to fix a clippy warning where `path` isn't being used.

tembo-operator/yaml/sample-message-queue.yaml
```diff
@@ -73,6 +73,8 @@ spec:
           - total_messages:
               description: Total number of messages that have passed into the queue.
               usage: GAUGE
+        target_databases:
```
ChuckHend, line 76: Do we also need to add this to the Stack definition at src/stacks/templates?

nhudson: No it will add it automatically when reconciling the configmap.

ChuckHend (approved):

PR #453, opened 2023-12-25
runs validate before running apply to find any validation errors right away.

shahadarsh:
Merry Christmas! 

* Calls `tembo validate` at the start of `tembo apply` to find any validation errors right away
* If `5432` port is taken it uses `5433` or `5434` to run a docker container for local context

### DIFF
--- a/tembo-cli/src/cli/docker.rs
+++ b/tembo-cli/src/cli/docker.rs
@@ -1,5 +1,6 @@
 use crate::Result;
 use anyhow::bail;
+use anyhow::Error;
 use simplelog::*;
 use spinners::{Spinner, Spinners};
 use std::process::Command as ShellCommand;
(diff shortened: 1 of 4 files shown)


tembo-cli/src/cli/docker.rs
```diff
@@ -41,24 +42,57 @@ impl Docker {
...
             );
             run_command(&command)?;
             sp.stop_with_message("- Docker Build & Run completed".to_string());
+            Ok(port)
         }
+    }
 
-        Ok(())
+    fn get_available_port() -> Result<i32> {
+        let ls_command = "docker ps -a --format '{{.Ports}}'";
```
vrmiguel, line 68: I'm wondering if we really need the `-a`

shahadarsh: Good catch. Not needed. Updated it.

sjmiller609 (approved):

PR #455, opened 2023-12-27
Tembo CLI apply improvements

sjmiller609:
- tembo apply delete + recreate if already exists
- tembo apply to local, always remove container when stopped
- tembo apply handle the case of container exists but stopped
- tembo apply can work with `tembo --verbose apply` to stream output from build + run to shell
- Include templates with build instead of download from github

### DIFF
--- a/tembo-cli/src/cli/docker.rs
+++ b/tembo-cli/src/cli/docker.rs
@@ -1,9 +1,11 @@
-use anyhow::bail;
+use std::io::{BufRead, BufReader};
+use anyhow::{bail, Context};
 use anyhow::Error;
 use simplelog::*;
 use spinners::{Spinner, Spinners};
-use std::process::Command as ShellCommand;
+use std::process::{Command as ShellCommand, Stdio};
 use std::process::Output;
+use std::thread;
 
 pub struct Docker {}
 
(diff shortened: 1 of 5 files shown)


tembo-cli/src/cli/docker.rs
```diff
@@ -153,18 +168,45 @@ impl Docker {
...
         .arg("-c")
         .arg(command)
-        .output()
-        .expect("failed to execute process");
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .with_context(|| format!("Failed to spawn command '{}'", command))?;
+
+    if verbose {
```
DarrenBaldwin07, line 180: `verbose` is a great addition

DarrenBaldwin07 (approved):

tembo-cli/src/cmd/apply.rs
```diff
@@ -476,23 +467,15 @@ pub fn get_rendered_dockerfile(
 pub fn get_rendered_migrations_file(
     instance_settings: HashMap<String, InstanceSettings>,
 ) -> Result<String, anyhow::Error> {
-    let filename = "migrations.sql.template";
-    let filepath =
-        "https://raw.githubusercontent.com/tembo-io/tembo/main/tembo-cli/tembo/migrations.sql.template";
+    // Include the migrations template directly into the binary
+    let contents = include_str!("../../tembo/migrations.sql.template");
```
DarrenBaldwin07, line 471: Much better!

PR #463, opened 2024-01-03
adds support for x86_64-apple, aarch64-apple, x86_64_linux, aarch64-linux

shahadarsh:
[Here](https://github.com/tembo-io/tembo/releases/tag/0.0.28) is an example release. It's set to pre-release as its only for testing. Release notes are not working. I think I know the reason and will fix it in the next PR.

Please ignore the number of commits. :)


.github/workflows/tembo_release.yml
```diff
@@ -12,26 +12,70 @@ jobs:
...
       - name: Check out the repo
         uses: actions/checkout@v4
 
+      - name: Install Deps Ubuntu
+        id: install_deps_ubuntu
+        run: |
+          set -x
+          apt update -y
+          apt install pkg-config libssl-dev -y
+        if: matrix.os == 'ubuntu-20.04'
```
theory, line 46: I suggest using `if: runner.os == 'Linux'` instead, so it should work without modification if you change the vision of Ubuntu.

.github/workflows/tembo_release.yml
```diff
@@ -12,26 +12,70 @@ jobs:
...
+      # TODO: Figure out a better way to fetch version for mac & linux
+      - name: CLI version Mac
+        working-directory: ./tembo-cli
+        id: cli_version_mac
+        run: |
+          set -x
+          cargo install cargo-get
+          TEMBO_CLI_NAME=$(cargo-get package.name)-$(cargo-get package.version)
+          echo "TEMBO_CLI_NAME=$TEMBO_CLI_NAME" >> $GITHUB_ENV
+        if: matrix.os == 'macos-latest'
```
theory, line 66:
```suggestion
        if: runner.os == 'macOS'
```

.github/workflows/tembo_release.yml
```diff
@@ -12,26 +12,70 @@ jobs:
...
+      - name: CLI version Ubuntu
+        working-directory: ./tembo-cli
+        id: cli_version_ubuntu
+        run: |
+          set -x
+          PKG_VERSION=$(awk -F ' = ' '$$1 ~ /version/ { gsub(/[\\"]/, "", $$2); printf("%s",$$2); exit; }' Cargo.toml)
+          TEMBO_VERSION=${PKG_VERSION#"version = "}
+          TEMBO_CLI_NAME=tembo-cli-${TEMBO_VERSION}
+          echo "TEMBO_CLI_NAME=$TEMBO_CLI_NAME" >> $GITHUB_ENV
+        if: matrix.os == 'ubuntu-20.04'
```
theory, line 77:
```suggestion
        if: runner.os == 'Linux'
```

theory (approved):
Can't say much about the build changes, but kibitzed a bit about using forward-compatible conditionals in your workflow. :-)

.github/workflows/tembo_release.yml
shahadarsh: Yup, good idea. I will change that in the next PR.

PR #466, opened 2024-01-04
implements overlay-configs options for tembo apply

joshuajerin:
- Adding --merge for overlay of Base and Second Instance toml files.

https://linear.app/tembo/issue/TEM-2162/tembo-apply-overlay-configs


joshuajerin:
Closes Linear-2162

tembo-cli/src/cli/tembo_config.rs
```diff
@@ -14,13 +14,13 @@ pub struct InstanceSettings {
     pub environment: String,
     pub instance_name: String,
     #[serde(default = "default_cpu")]
-    pub cpu: String,
+    pub cpu: Option<String>,
```
sjmiller609, line 17: maybe we don't need to change this one since we ended up with the new struct

tembo-cli/src/cmd/apply.rs
```diff
@@ -30,16 +31,20 @@ use crate::cli::file_utils::FileUtils;
...
 /// Deploys a tembo.toml file
 #[derive(Args)]
-pub struct ApplyCommand {}
+pub struct ApplyCommand {
+    #[clap(long, short = 'm')]
+    pub merge: Option<String>,
+}
 
-pub fn execute(verbose: bool) -> Result<(), anyhow::Error> {
+pub fn execute(verbose: bool, _merge_path: Option<String>) -> Result<(), anyhow::Error> {
```
sjmiller609, line 47: accidentally added parameter

sjmiller609 (changes requested):
Great change, and thank you! Let's add a test that uses --merge option

joshuajerin (commented):
Done!

tembo-cli/tests/tomls/merge/second-instance.toml
```diff
@@ -0,0 +1,5 @@
+[instance]
```
sjmiller609, line 1: Can you please remove all but 1 settings in the overlay (this) file, I recommend "instance_name" as the only setting to have in this file. Also it might be more clear to name the file "overlay.toml" or "merge.toml"

sjmiller609 (changes requested):
Looking great, just a few things. I think it looks like maybe the overlay will reset the instance settings to default if not specified in the overlay config, so I want to see a test for that. Also there are some files that don't need to be committed that could be cleaned up. To get the lint rules to pass in CI, run `cargo clippy --fix --all` and run `cargo fmt`.

tembo-cli/src/cmd/apply.rs
```diff
@@ -434,26 +439,69 @@ fn get_trunk_installs(
...
+    }
+}
 
-    let contents = match fs::read_to_string(file_path.clone()) {
-        Ok(c) => c,
-        Err(e) => {
-            panic!("Couldn't read context file {}: {}", file_path, e);
-        }
-    };
+fn merge_settings(base: &InstanceSettings, overlay: OverlayInstanceSettings) -> InstanceSettings {
```
DarrenBaldwin07, line 457: We should take a look at this crate to help handle merging potentially - maybe @vrmiguel has opinions here too? https://crates.io/crates/serde-toml-merge

tembo-cli/src/cmd/apply.rs
```diff
@@ -41,28 +43,34 @@ const POSTGRESCONF_NAME: &str = "postgres.conf";
...
 
-pub fn execute(verbose: bool) -> Result<(), anyhow::Error> {
+pub fn execute(verbose: bool, _merge_path: Option<String>) -> Result<(), anyhow::Error> {
     info!("Running validation!");
     super::validate::execute(verbose)?;
     info!("Validation completed!");
 
     let env = get_current_context()?;
 
+    let instance_settings = get_instance_settings(_merge_path.clone())?;
```
shahadarsh, line 58: This call feels unnecessary since `get_instance_settings` is already called in the `execute_docker` & `execute_tembo_cloud` functions.

tembo-cli/src/cmd/apply.rs
```diff
@@ -41,28 +43,34 @@ const POSTGRESCONF_NAME: &str = "postgres.conf";
...
+    println!("Instance settings: {:?}", instance_settings);
```
shahadarsh, line 59: You can add a println in the `get_instance_settings` function towards the end saying overlay file merged successfully.  I don't think you need to print the values but I'm fine keeping it if you think it will be helpful and just move it to `get_instance_settings` function.

tembo-cli/src/cmd/delete.rs
```diff
@@ -32,7 +30,7 @@ pub fn execute() -> Result<(), anyhow::Error> {
 }
 
 fn execute_tembo_cloud(env: Environment) -> Result<(), anyhow::Error> {
-    let instance_settings: HashMap<String, InstanceSettings> = get_instance_settings()?;
+    let instance_settings = get_instance_settings(None)?;
```
shahadarsh, line 33: Since you can't overlay `instance_name` I think passing None here should be fine.

tembo-cli/src/main.rs
```diff
@@ -52,8 +52,8 @@ fn main() -> Result<(), anyhow::Error> {
         SubCommands::Init(_init_cmd) => {
             init::execute()?;
         }
-        SubCommands::Apply(_apply_cmd) => {
-            apply::execute(app.global_opts.verbose)?;
+        SubCommands::Apply(ref _apply_cmd) => {
```
shahadarsh, line 55: Not sure if we need to pass this as ref. Lets try it out together over a call.

tembo-cli/tests/tomls/merge/.gitignore
```diff
@@ -0,0 +1,3 @@
+postgres.conf
```
shahadarsh, line 1: Great.

tembo-cli/tests/tomls/merge/.gitignore
shahadarsh, line 1: Thanks for creating the .gitignore.

shahadarsh (approved):
Great work @joshuajerin. Not an easy task for your 1st one. I did some basic testing and local & cloud work as expected. There are a few other improvements we can do especially in how the merge logic is written but I'm fine merging this and doing a follow up PR with improvements.

PR #470, opened 2024-01-05
compresses release binaries

shahadarsh:
* compresses release binaries
* uses runner.os instead of matrix.os 
* uses different changelog action


.github/workflows/tembo_release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           set -x
           apt update -y
           apt install pkg-config libssl-dev -y
-        if: matrix.os == 'ubuntu-20.04'
```
shahadarsh, line 46: @theory Thanks for the comment in the previous PR. I changed these.

sjmiller609 (approved):

PR #471, opened 2024-01-06
Add validation of configuration values from the tembo.toml file

nhudson:
This change will add configuration validation against the `temboclient` for the following items.  The validation will fail if the client cannot validate the values.
- environment
- cpu
- memory
- replicas
- storage
- stack types

Added tests using a crate called [rstest](https://github.com/la10736/rstest) that allows you to easily add macros for fixture based tests.  If someone would rather me write all the test scenarios out instead I can, just comment and let me know.


shahadarsh (approved):
LGTM :thumbsup:

Thanks for doing this. The tests are failing because it says CPU value is invalid for one of the tests but I think its a valid value. I can take a look at it later. I might have to generate the rut client again with latest values.

sjmiller609 (approved):

PR #475, opened 2024-01-09
fix fractional cpu parsing

ChuckHend:


sjmiller609 (approved):
A non-deterministic result would result in the workload restarting over and over, but I think the test covers that situation

PR #476, opened 2024-01-09
Adding Tembo logs for docker

joshuajerin:


tembo-cli/src/main.rs
```diff
@@ -61,6 +63,9 @@ fn main() -> Result<(), anyhow::Error> {
         SubCommands::Delete(_delete_cmd) => {
             delete::execute()?;
         }
+        SubCommands::Logs(logs_cmd) => {
```
DarrenBaldwin07, line 66: Likely can remove or set as unused for `logs_cmd`

tembo-cli/src/cmd/logs.rs
```diff
@@ -0,0 +1,117 @@
...
+pub fn execute() -> Result<()> {
+    let instance_settings = super::apply::get_instance_settings(None)?;
+    let mut errors = Vec::new();
+    let env = get_current_context()?;
+
+    if env.target == Target::Docker.to_string() {
+        for (instance_name, _settings) in instance_settings {
+            match docker_logs(&instance_name) {
+                Ok(_) => (),
+                Err(e) => errors.push((instance_name, e)),
```
DarrenBaldwin07, line 21: What are we doing with this `errors` vec? Seems liek we should log the errors at some point

joshuajerin: we are printing out the logs too in docker_logs fn
```
let logs_stdout = String::from_utf8_lossy(&output.stdout);
    let logs_stderr = String::from_utf8_lossy(&output.stderr);
```

tembo-cli/src/main.rs
joshuajerin: Removed it!

PR #481, opened 2024-01-10
Adding logs for cloud

joshuajerin:


tembo-cli/src/cmd/logs.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        let instance_id = if let Some(id) = instance_id_option {
+            id
+        } else {
+            eprintln!("Instance ID not found for {}", value.instance_name);
+            continue;
+        };
+
+        let query = format!("{{tembo_instance_id=\"{}\"}}", instance_id);
+        let url = "https://api.data-1.use1.tembo.io/loki/api/v1/query_range";
```
sjmiller609, line 91: Can you please read this domain name from the profile file so that it can work in either dev or prod cloud, for example

tembo-cli/src/cmd/logs.rs
```diff
@@ -0,0 +1,108 @@
...
+use clap::Args;
+use reqwest::blocking::Client;
+use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION};
+use serde::{Deserialize, Serialize};
+use temboclient::apis::configuration::Configuration;
+
+#[derive(Args)]
+pub struct LogsCommand {
+    #[clap(short, long)]
+    pub verbose: bool,
```
sjmiller609, line 13: I think logs command should not have verbose as a field since that's already a global option, but I may be mistaken about that.

sjmiller609 (commented):
this is looking great! also adding some tests would be nice

tembo-cli/src/cmd/logs.rs
joshuajerin: You're right. It shouldn't have it.

tembo-cli/src/cmd/logs.rs
joshuajerin: Done!

sjmiller609 (approved):

PR #483, opened 2024-01-12
More `tembo-cli` output polish

DarrenBaldwin07:
## Fixes + adds a few polish items after syncing with @shahadarsh 
- More overall output polish in all commands
- Fixes bug where the `tembo apply` loader would show after an error for a few seconds


shahadarsh (approved):
LGTM :thumbsup:

shahadarsh:
@DarrenBaldwin07 We should bump the version in Cargo.toml.

PR #484, opened 2024-01-16
adds support for running multiple instances in tembo.toml file using docker-compose & traefik

shahadarsh:
* adds support for running multiple instances in tembo.toml file using docker-compose & traefik
  * Example file [here](https://github.com/tembo-io/tembo/blob/53c70d3995f79155cc27fa07fbb7fb47b9229419/tembo-cli/tests/tomls/multiple-instances/tembo.toml)
* uses @sjmiller609's example from [here](https://github.com/tembo-io/tembo/tree/main/tembo-cli/tests/local-app-services) for the docker-compose template
* regenerates rust client from control plane

Output for `tembo apply` against local context:

```➜  tembo-toml [dev-eks] tembo apply
✓ Configuration is valid
✓ Docker Build completed for test-via-cli-69
✓ Docker Build completed for test-via-cli-mobile-3
✓ Docker Compose Up completed
✓ SQL migration completed for test-via-cli-69

┃ >>> T E M B O local instance started:

 ➜ Url: postgres://postgres:postgres@test-via-cli-69.local.tembo.io:5432
 ➜ OLTP

✓ SQL migration completed for test-via-cli-mobile-3

┃ >>> T E M B O local instance started:

 ➜ Url: postgres://postgres:postgres@test-via-cli-mobile-3.local.tembo.io:5432
 ➜ Standard

### CODE REVIEW

File: tembo-cli/tests/tomls/multiple-instances/tembo.toml [hunk 1]
```diff
@@ -0,0 +1,16 @@
+[defaults]
```
sjmiller609, line 1: this should be named something other than defaults, since defaults is supposed to mean defaults for all instances, not the name of an instance

sjmiller609 (approved):

PR #485, opened 2024-01-17
--set

joshuajerin:
Closes Linear-2724

We are adding --set subcommand inside apply to temporarily set user's specifying setting to that value.

To test it out locally, you can run "tembo apply --set (instance name).(instance setting) = (instance value).
There is a folder inside Tembo-cli/examples/set for a example tembo.toml file

When tested locally, it changes the stack_type from "Standard" to "OLTP" in Docker as seen below.
Screenshot 2024-01-22 at 11 21 55

When tested in cloud, we give the command to change CPU from "1" to "2" as seen below with a Screenshot from mahout.
Screenshot 2024-01-22 at 11 26 25
Screenshot 2024-01-22 at 11 29 26


tembo-cli/src/cmd/apply.rs
```diff
@@ -46,28 +46,64 @@ const POSTGRESCONF_NAME: &str = "postgres.conf";
...
+    if key_parts.len() != 2 {
+        println!("Error: Invalid format (missing '.')");
+        return Err(Error::msg("Invalid format for --set"));
+    }
+
+    let instance_name = key_parts[0].to_string();
+    let setting_name = key_parts[1].to_string();
+    let setting_value = parts[1].to_string();
+
+    println!("Instance Name: {}", instance_name);
```
shahadarsh, line 92: Do we really need to print these here?

tembo-cli/src/cmd/apply.rs
```diff
@@ -124,8 +160,9 @@ fn execute_docker(verbose: bool, _merge_path: Option<String>) -> Result<(), anyh
 pub fn execute_tembo_cloud(
     env: Environment,
     _merge_path: Option<String>,
+    _set_arg: Option<String>,
 ) -> Result<(), anyhow::Error> {
-    let instance_settings = get_instance_settings(_merge_path)?;
+    let instance_settings = get_instance_settings(_merge_path, None)?;
```
shahadarsh, line 165: Should this be `None`?  or you need to pass `_set_arg` to this?

tembo-cli/src/cmd/apply.rs
```diff
@@ -501,6 +539,28 @@ pub fn get_instance_settings(
         }
     }
 
+    if let Some(set_arg) = set_arg {
```
shahadarsh, line 542: Can you move this to a separate function with an appropriate name that you can call from here? Smaller functions the better.

shahadarsh: I would also move the overlay logic in this function to a separate function.

shahadarsh:
@joshuajerin Great progress. A few points: 
 
* Can you add a test similar to the one you added for merge?
* In the PR description, can you explain the steps you took to test it locally and against the cloud? If you can specify the outputs there that would be great. 
* Because of my merge earlier today you will have to fix the merge conflicts (sorry). You will see I moved the tests from `main.rs` to `apply.rs` because it's testing a function in `apply.rs`

tembo-cli/examples/merge/overlay.toml
```diff
@@ -1,4 +1,4 @@
-[merge]
-environment = "prod"
+[defaults]
```
shahadarsh, line 1: @joshuajerin Can you keep this as `merge` instead of `defaulta`. This was a PR comment Steven provided so I made those changes in [this](https://github.com/tembo-io/tembo/pull/487) PR

tembo-cli/tests/tomls/set/tembo.toml
```diff
@@ -0,0 +1,9 @@
+[defaults]
```
shahadarsh, line 1: Can you use a different name here? This is for the same reason mentioned above. Also, can you move this under the examples directory?

PR #499, opened 2024-01-23
adding --merge and --set flags to docs

joshuajerin:


DarrenBaldwin07:
We likely dont have to maintain this doc if its also being mirrored in https://tembo.io/docs

PR #510, opened 2024-01-25
Upgrade to actions/checkout@v4

theory:


nhudson:
This will most likely break the jobs. I've seen v4 not work at all.

theory:
Huh. Has worked like a champ for me. v3 is deprecated (nodjs is too old).

theory:
@nhudson I reverted the changes in 7bea6cb to see what would happen. Exactly the same tests fail as with the changes, in 9932612. I think perhaps something else needs fixing.

ChuckHend (approved):

nhudson (approved):

PR #511, opened 2024-01-25
Adding tembo logs for Docker

joshuajerin:


sjmiller609 (approved):
Good after fixing lints

PR #512, opened 2024-01-25
improved error handling in trunk meta handler

ChuckHend:
Intention is to not panic when we encounter an extension that is not in trunk. For example, when an extension is built into the base image and not available in trunk.


ianstanton (approved):
:call_me_hand:

PR #525, opened 2024-02-03
Updated Hydra version for olap stack

shhnwz:


ChuckHend (approved):
Will need to increase the version in cargo.toml when it's ready to be released.

nhudson (approved):

PR #527, opened 2024-02-05
Add single-instance example for CLI init

ianstanton:
The `tembo init` command relies on https://raw.githubusercontent.com/tembo-io/tembo/main/tembo-cli/examples/single-instance/tembo.toml, which was removed in this PR: 
- https://github.com/tembo-io/tembo/pull/487

Now `tembo init` produces a `tembo.toml` file with contents `404: Not Found`. This PR re-adds the file so it can be used for `tembo init`.

### DIFF
--- a/tembo-cli/examples/single-instance/tembo.toml
+++ b/tembo-cli/examples/single-instance/tembo.toml
@@ -1,0 +1,34 @@
+# Tembo.toml configuration file version
+# version = "1.0.0"
+
+# Default settings for instances
+[defaults]
+environment = "dev"
+instance_name = "test-via-cli-35"
+cpu = "1"
+memory = "2Gi"
+storage = "10Gi"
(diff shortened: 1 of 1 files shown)


ianstanton:
I added the same file that was there before: https://github.com/tembo-io/tembo/blob/72083bff5670fe9aaaf0977598b78ca2ee5ddd43/tembo-cli/examples/single-instance/tembo.toml

shahadarsh:
> I added the same file that was there before: https://github.com/tembo-io/tembo/blob/72083bff5670fe9aaaf0977598b78ca2ee5ddd43/tembo-cli/examples/single-instance/tembo.toml

Just updated a few small things but looks good otherwise.

shahadarsh (approved):
LGTM :thumbsup:

PR #528, opened 2024-02-06
Drop Traefik CRD dependency from `tembo-operator`

ianstanton:
This PR adds more flexibility for using the Tembo Operator with an ingress controller other than Traefik. This logic was already present in the main controller, but missing form the `AppServices` controller:
- When `DATA_PLANE_BASEDOMAIN` environment variable is not set, skip anything related to `IngressRouteTCP` and other Traefik specific custom resources
- When `USE_SHARED_CA` is not set, skip certs volume mount in `AppServices` deployment. This is not related to Traefik CRDs, but necessary for the OSS tembo operator experience.

I am leaving a test out of this PR for a few reasons:
- In order to truly test that we can create instances when traefik CRDs are not present on a cluster, we'd need to run tests on a cluster _without_ the CRDs installed. They're present on the cluster we currently run tests on. I think some day we may want a separate test suite for the OSS user experience, but it seems a bit overkill today.
- I had also considered running a test with `DATA_PLANE_BASEDOMAIN` and `USE_SHARED_CA` unset. Because the tests run in parallel, we'd need to run this specific test after all others had passed. I couldn't find a clean way to do this at the moment.


tembo-operator/src/app_service/manager.rs
```diff
@@ -719,57 +732,60 @@ pub async fn reconcile_app_services(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(
...
-        }
-    }
-
-    for appsvc in appsvcs.iter() {
-        let app_name = appsvc.name.clone();
-
-        match reconcile_ingress_tcp(
+    // If DATA_PLANE_BASEDOMAIN is not set, skip ingress reconciliation
+    if domain.is_some() {
+        match reconcile_ingress(
```
ianstanton, line 737: Move into conditional

ChuckHend (approved):

PR #529, opened 2024-02-06
moves app_services related code to tembo-stacks crate from control plane so it can be used at other places

shahadarsh:
* Moves `apps` folder from `control-plane` to `tembo-stacks` crate so that `tembo-cli` and in future other places can use this crate instead of repeating the logic
* In future, we can also move the common logic that `tembo-operator` has into this crate.

Related PR: https://github.com/tembo-io/control-plane/pull/509


ChuckHend (approved):
🚀

vrmiguel (approved):

DarrenBaldwin07 (approved):

PR #530, opened 2024-02-07
conductor/stacks: set images to use for Pg14 and Pg16

vrmiguel:
New fields:

```yml
# Within the Docker image registry used (quay.io or ECR), this is the
# organization/username the images below are contained in
organization: tembo
# Image names to use for each Pg version
# Note: images for Pg15 were kept as is as to preserve full backwards compatibility
images:
  14: "standard-cnpg:pg14-f1721c2"
  15: "standard-cnpg:15.3.0-1-3d9a853"
  16: "standard-cnpg:pg16-f1721c2"
  ```

Control Plane will use the above information to set the instance's image according to the `pg_version` attribute in the `POST /instances` endpoint (defaults to Pg15)

### DIFF
--- a/tembo-operator/src/stacks/templates/api.yaml
+++ b/tembo-operator/src/stacks/templates/api.yaml
@@ -3,7 +3,7 @@ description: Tembo Stack with REST and graphQL interfaces.
 organization: tembo
 images:
   14: "standard-cnpg:14-a0a5ab5"
-  15: "standard-cnpg:15.3.0-1-3d9a853"
+  15: "standard-cnpg:15-a0a5ab5"
   16: "standard-cnpg:16-a0a5ab5"
 stack_version: 0.1.0
 appServices:
--- a/tembo-operator/src/stacks/templates/api.yaml
+++ b/tembo-operator/src/stacks/templates/api.yaml
@@ -1,5 +1,6 @@
 name: API
 description: Tembo Stack with REST and graphQL interfaces.
+repository: "quay.io/tembo"
 organization: tembo
 images:
   14: "standard-cnpg:14-a0a5ab5"
--- a/tembo-operator/src/stacks/templates/api.yaml
+++ b/tembo-operator/src/stacks/templates/api.yaml
@@ -2,9 +2,9 @@ name: API
 description: Tembo Stack with REST and graphQL interfaces.
 organization: tembo
 images:
-  14: "standard-cnpg:pg14-f1721c2"
+  14: "standard-cnpg:14-a0a5ab5"
   15: "standard-cnpg:15.3.0-1-3d9a853"
-  16: "standard-cnpg:pg16-f1721c2"
+  16: "standard-cnpg:16-a0a5ab5"
 stack_version: 0.1.0
 appServices:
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: tembo-operator/src/stacks/templates/api.yaml [hunk 1]
```diff
@@ -1,7 +1,10 @@
 name: API
 description: Tembo Stack with REST and graphQL interfaces.
-repository: "quay.io/tembo"
```
nhudson, line 3: You might want to add back the repository section.  Even though there is a default, I think leaving this here is necessary.

nhudson, line 3: You might want to add back the repository section. Even though there is a default, I think leaving this here is necessary.

tembo-operator/src/stacks/templates/api.yaml
```diff
@@ -1,7 +1,10 @@
...
-image: "standard-cnpg:15.3.0-1-3d9a853"
+organization: tembo
+images:
+  14: "standard-cnpg:pg14-f1721c2"
+  15: "standard-cnpg:15.3.0-1-3d9a853"
```
nhudson, line 6: The PG 15 image name needs to be like the others.  We now publish it as `*-cnpg:pg<pg version>-<git sha>`

tembo-operator/src/stacks/templates/standard.yaml
```diff
@@ -1,7 +1,10 @@
 name: Standard
 description: A balanced Postgres instance optimized for OLTP workloads.
-repository: "quay.io/tembo"
-image: "standard-cnpg:15.3.0-1-3d9a853"
+organization: tembo
+images:
+  14: "standard-cnpg:pg14-f1721c2"
+  15: "standard-cnpg:15.3.0-1-3d9a853"
+  16: "standard-cnpg:pg16-f1721c2"
```
nhudson, line 7: These are still pointing to the incorrect image tags

tembo-operator/src/stacks/templates/oltp.yaml
```diff
@@ -1,7 +1,10 @@
 name: OLTP
 description: A Postgres instance optimized for OLTP workloads.
-repository: "quay.io/tembo"
-image: "standard-cnpg:15.3.0-1-3d9a853"
+organization: tembo
+images:
+  14: "standard-cnpg:14-a0a5ab5"
+  15: "standard-cnpg:15.3.0-1-3d9a853"
+  16: "standard-cnpg:16-a0a5ab5"
```
nhudson, line 7: Lets make these all the same tag

```
 14: "standard-cnpg:14-a0a5ab5"
 15: "standard-cnpg:15-a0a5ab5"
 16: "standard-cnpg:16-a0a5ab5"
 ```

nhudson (changes requested):

tembo-operator/src/stacks/templates/standard.yaml
vrmiguel: Oops, good catch. Fixed

nhudson (approved):
Approved, just don't forget to go back and fix the data warehouse images.

PR #534, opened 2024-02-08
uses tembo-stacks crate for merge logic between stacks, instance, app…

shahadarsh:
…_services

### DIFF
--- a/tembo-cli/src/cmd/apply.rs
+++ b/tembo-cli/src/cmd/apply.rs
@@ -2,6 +2,10 @@ use anyhow::Context as AnyhowContext;
 use anyhow::Error;
 use clap::Args;
 use colorful::Colorful;
+use controller::app_service::types::AppService;
+use controller::extensions::types::Extension as ControllerExtension;
+use controller::extensions::types::ExtensionInstallLocation as ControllerExtensionInstallLocation;
+use controller::extensions::types::TrunkInstall as ControllerTrunkInstall;
 use controller::stacks::get_stack;
 use controller::stacks::types::StackType as ControllerStackType;
 use log::info;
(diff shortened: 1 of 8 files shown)


tembo-cli/src/cmd/apply.rs
```diff
@@ -179,13 +195,43 @@ fn docker_apply(
...
+        extensions,
+        trunk_installs,
+        app_services,
+        pg_configs: _,
+    } = merge_app_reqs(
+        instance_settings.app_services.clone(),
+        stack.app_services.clone(),
+        extensions,
+        trunk_installs,
+        Option::None,
```
shahadarsh, line 231: Skipping the `pg_config` merge here and using the existing logic that's there

tembo-cli/tembo/docker-compose.yml.template
```diff
@@ -17,27 +17,55 @@ services:
       - "traefik.tcp.routers.{{instance.instance_name}}.entrypoints=postgresql"
       - "traefik.tcp.routers.{{instance.instance_name}}.tls.passthrough=true"
       - "traefik.tcp.services.{{instance.instance_name}}.loadbalancer.server.port=5432"
+{% if instance.controller_app_services -%}
+{% for key, app_services in instance.controller_app_services -%}
 
-  {% if instance.app_services.restapi %}
-  {{instance.instance_name}}-postgrest:
-    image: postgrest/postgrest:v10.0.0
+{% set_global pathPrefixes = "" -%}
```
shahadarsh, line 23: The template is getting complex. I will look at simplifying this either by moving it to rust or figuring out another way in the coming PRs.

tembo-cli/src/cmd/apply.rs
```diff
@@ -528,6 +574,32 @@ fn get_extensions(
     vec_extensions
 }
 
+fn get_extensions_controller(
```
shahadarsh, line 577: Now that we have `tembo-stacks` more types can be reused but will require significant work in CLI so need to convert it to use it with `merge_app_reqs`.

tembo-cli/tembo/docker-compose.yml.template
sjmiller609: I agree, at this point it seems like we should just have a Value type constructed in rust then formatted into the yaml file

sjmiller609 (approved):

PR #536, opened 2024-02-09
Bump chart to include traefik dependency fix

ianstanton:
Bump tembo-operator helm chart to include #528


ianstanton:
This isn't necessary because the helm chart will use the latest image. Closing.

PR #537, opened 2024-02-09
Add App Definition for `pganalyze`

ianstanton:
Add app definition for `pganalyze` and sample YAML for OSS users.


tembo-operator/yaml/sample-pganalyze.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: coredb.io/v1alpha1
```
ianstanton, line 1: Sample for OSS users

tembo-stacks/src/apps/pganalyze.yaml
```diff
@@ -0,0 +1,9 @@
+name: !pganalyze
+appServices:
+  - image: quay.io/pganalyze/collector:v0.53.0
+    name: pganalyze-collector
+    env:
+      - name: DB_URL
+        valueFromPlatform: ReadWriteConnection
+      - name: PGA_API_KEY
+        value: ""
```
ianstanton, line 9: @ChuckHend We're going to be collecting the `PGA_API_KEY` value from users when they activate this app. I'm assuming we'll be able to pass that down through here. Is that true? Is it appropriate to set that value as `""` in this template?

ChuckHend: Yes the value can be passed down, but it will need to be handled in the control-plane. Nothing in the operator requires the env var to be declared here as empty. So I would say only set this unless the application requires it. 

OSS users will set that directly in the spec. In Tembo cloud thats handled in the control-plane.

ianstanton: This one is required by the app, so maybe we leave it here as an indicator / reminder that it needs to be overridden by control plane

ChuckHend: It might be easier to troubleshoot shoot if this is not provided rather than being assigned an empty string. Just thinking about how those logs will end up looking. There's no way to differentiate user not providing a value vs user providing an empty value, the way this is currently set up.

ChuckHend (approved):
Recommend removing the empty string assignment for the api key env var.

PR #538, opened 2024-02-12
Added Columnar parallel processes parameter

shhnwz:


tembo-operator/src/stacks/config_engines.rs
```diff
@@ -79,6 +84,8 @@ pub fn olap_config_engine(stack: &Stack) -> Vec<PgConfig> {
     let max_parallel_workers = olap_max_parallel_workers(vcpu);
     let max_parallel_workers_per_gather = olap_max_parallel_workers_per_gather(vcpu);
     let max_worker_processes = olap_max_worker_processes(vcpu);
+    let effective_io_concurrency: i32 = 100;
```
ChuckHend, line 87: move this to a constant, similar to how we have `const DEFAULT_MAINTENANCE_WORK_MEM_MB: i32 = 64;` defined.

shhnwz: Introduced const DEFAULT_EFFECTIVE_IO_CONCURRENCY: i32 = 100.

ChuckHend (approved):

PR #539, opened 2024-02-12
Tembologin

joshuajerin:


tembo-cli/src/cmd/login.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    HttpResponse::Ok().body(format!("Token received and saved: {}", token))
+}
+
+async fn start_server(notify: Arc<Notify>) -> Result<()> {
+    let notify_data = web::Data::new(notify.clone());
+
+    let server = HttpServer::new(move || {
+        let cors = Cors::default()
+            .allowed_origin("https://pr-462.dkwugv8ig5mdu.amplifyapp.com")
```
sjmiller609, line 68: You can please remove this line before merging

tembo-cli/src/cmd/login.rs
```diff
@@ -0,0 +1,116 @@
...
+    let lifetime = lifetime.trim().to_string();
+
+    Ok(lifetime)
+}
+
+fn save_token_to_file(token: &str) -> Result<(), Error> {
+    let home_dir = dirs::home_dir().expect("Could not find home directory");
+    let credentials_path = home_dir.join(".tembo/credentials");
+
+    let new_contents = format!(
```
sjmiller609, line 106: I think this ought to save the token into the currently selected profile’s corresponding credentials. In this current way, it will overwrite the content, but the user may have configurations here like multiple different credentials contexts, so overwriting the while file might be annoying to the user.

sjmiller609: So I think the login command should not assume any of these api endpoints, instead it should modify the existing file to add just the token.

sjmiller609 (approved):
Looks good, please address two minor requests then good to merge if CI is green.

sjmiller609 (approved):
I think maybe some changes needed to have it work against anything besides prod cloud because the prod cloud domains are coded in

tembo-cli/src/cmd/login.rs
```diff
@@ -0,0 +1,137 @@
...
+use std::sync::Arc;
+use tokio::sync::Notify;
+use tokio::time::{self, Duration};
+use webbrowser;
+
+#[derive(Args)]
+pub struct LoginCommand {}
+
+#[derive(Deserialize)]
+struct TokenResponse {
```
DarrenBaldwin07, line 17: We should rename this to `TokenBody` or `TokenRequest` since this is a request body type and not really a response

DarrenBaldwin07 (approved):
Looks great!

PR #546, opened 2024-02-14
remove app role and db from being created

nhudson:
Remove unnecessary `app` role and database from being created.

Fixes: [TEM-3094](https://linear.app/tembo/issue/TEM-3094/consider-dropping-creation-of-the-app-database)


nhudson:
Removing it from mahout as well https://github.com/tembo-io/mahout/pull/482

ianstanton (approved):
🎉

PR #551, opened 2024-02-16
uses sqlx query instead of migrations for creating extensions

shahadarsh:
* uses sqlx query instead of migrations for creating extensions
* removes migrations for tembo cloud

### DIFF
--- a/tembo-cli/src/cmd/apply.rs
+++ b/tembo-cli/src/cmd/apply.rs
@@ -9,6 +9,8 @@ use controller::extensions::types::TrunkInstall as ControllerTrunkInstall;
 use controller::stacks::get_stack;
 use controller::stacks::types::StackType as ControllerStackType;
 use log::info;
+use spinoff::spinners;
+use spinoff::Spinner;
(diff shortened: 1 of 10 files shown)


tembo-cli/src/cmd/apply.rs
```diff
@@ -165,18 +157,27 @@ fn docker_apply(
...
-        ))?;
+    for (_key, instance_setting) in final_instance_settings.clone().iter() {
+        let instance_name = &instance_setting.instance_name;
+
+        let mut sp = Spinner::new(spinners::Dots, "Creating extensions", spinoff::Color::White);
+
+        for ext in instance_setting.final_extensions.clone().unwrap().iter() {
+            let query = &format!("CREATE EXTENSION IF NOT EXISTS {} CASCADE", ext.name);
+
+            Runtime::new().unwrap().block_on(SqlxUtils::execute_sql(
```
shahadarsh, line 168: TODO: Check if there is a better way to run this later.

sjmiller609 (approved):

PR #564, opened 2024-02-22
cli: don't quote Postgres configs in instance creation

vrmiguel:


sjmiller609 (approved):

shahadarsh:
@vrmiguel I'm gonna merge it and make changes on top of this since it does remove the quotes. It doesn't fix the issue but it's still better than it was previously.

PR #567, opened 2024-02-22
fetches loadable_libraries using trunk api

shahadarsh:
I'm doing more testing on this mainly around `priority` and will do a follow-up PR with any improvements.


tembo-cli/src/cmd/apply.rs
```diff
@@ -762,17 +766,21 @@ pub fn get_rendered_dockerfile(
...
+    let mut shared_preload_libraries: Vec<Library> = Vec::new();
 
     if let Some(ps_config) = postgres_configs {
         for p_config in ps_config.clone().into_iter() {
             match p_config.name.as_str() {
                 "shared_preload_libraries" => {
-                    shared_preload_libraries.push(p_config.value.to_string());
+                    shared_preload_libraries.push(Library {
+                        name: p_config.value.to_string(),
+                        priority: MAX_INT32,
```
shahadarsh, line 782: I don't think there is any way to get priority for these so setting them to `2147483647`.

shahadarsh: This is when `shared_preload_libraries` is specified in stack, app or directly for an instance.

tembo-cli/src/cmd/apply.rs
```diff
@@ -823,10 +834,17 @@ fn get_postgres_config(
         }
     }
 
-    let config = shared_preload_libraries
+    let final_lodable_libs = Runtime::new().unwrap().block_on(get_loadable_libraries(
+        shared_preload_libraries.clone(),
+        extensions,
+    ));
+
+    let config = final_lodable_libs
```
sjmiller609, line 842: Typo “lodable”

tembo-cli/src/cmd/apply.rs
```diff
@@ -823,10 +834,17 @@ fn get_postgres_config(
...
+        .unwrap()
```
sjmiller609, line 843: I’m not sure is this unwrap definitely safe or does it need error handling?

tembo-cli/src/cmd/apply.rs
```diff
@@ -838,6 +856,41 @@ fn get_postgres_config(
...
+        let trunk_projects_url =
+            "https://registry.pgtrunk.io/api/v1/trunk-projects?extension-name=";
+
+        for ext in extensions.iter() {
+            let response = reqwest::get(format!("{}{}", trunk_projects_url, ext.name))
+                .await?
+                .text()
+                .await?;
+
+            let mut trunk_projects: Vec<TrunkProject> = serde_json::from_str(&response).unwrap();
```
sjmiller609: Unhandled error case here with unwrap

tembo-cli/src/cmd/apply.rs
```diff
@@ -838,6 +856,41 @@ fn get_postgres_config(
...
 
+async fn get_loadable_libraries(
+    mut shared_preload_libraries: Vec<Library>,
+    maybe_extensions: Option<&Vec<ControllerExtension>>,
+) -> Result<Vec<Library>, anyhow::Error> {
+    if let Some(extensions) = maybe_extensions {
+        let trunk_projects_url =
+            "https://registry.pgtrunk.io/api/v1/trunk-projects?extension-name=";
+
+        for ext in extensions.iter() {
```
sjmiller609, line 867: I’m not sure but I think this logic is not handling the case of other trunk projects with matching extension names. Some different trunk projects have extensions of the same name

sjmiller609 (changes requested):

tembo-cli/src/cmd/apply.rs
sjmiller609: I’m saying when you look up by extension name, you might also get other unrelated trunk projects with extensions of the same name

tembo-cli/src/cmd/apply.rs
```diff
@@ -838,6 +856,41 @@ fn get_postgres_config(
...
+            let response = reqwest::get(format!("{}{}", trunk_projects_url, ext.name))
+                .await?
+                .text()
+                .await?;
+
+            let trunk_projects: Vec<TrunkProject> = serde_json::from_str(&response).unwrap();
```
sjmiller609, line 873: also I think this unwrap looks unsafe, if the response from the API does not fit the TrunkProject struct for some reason

tembo-cli/src/cmd/apply.rs
shahadarsh: Thanks. Added error handling.

tembo-cli/src/cmd/apply.rs
shahadarsh: Done.

tembo-cli/src/cmd/apply.rs
shahadarsh: Please check the latest commit to see if it makes sense to you.

sjmiller609 (approved):

PR #578, opened 2024-02-24
Add check to make sure VolumeSnapshotContent is created and ready

nhudson:
* Adding check to ensure the `VolumeSnapshotContent` is ready and applied before applying the `VolumeSnapshot`

This hopefully fixes a possible race condition where  the snapshot cannot be bound to the snapshotcontents.

* minor fixes and cleanup

### DIFF
--- a/tembo-operator/src/snapshots/volumesnapshots.rs
+++ b/tembo-operator/src/snapshots/volumesnapshots.rs
@@ -3,10 +3,12 @@ use crate::{
     snapshots::{
         volumesnapshotcontents_crd::{
             VolumeSnapshotContent, VolumeSnapshotContentDeletionPolicy,
-            VolumeSnapshotContentSource, VolumeSnapshotContentSpec,
+            VolumeSnapshotContentSource, VolumeSnapshotContentSpec, VolumeSnapshotContentStatus,
             VolumeSnapshotContentVolumeSnapshotRef,
         },
-        volumesnapshots_crd::{VolumeSnapshot, VolumeSnapshotSource, VolumeSnapshotSpec},
+        volumesnapshots_crd::{
+            VolumeSnapshot, VolumeSnapshotSource, VolumeSnapshotSpec, VolumeSnapshotStatus,
+        },
     },
     Context,
 };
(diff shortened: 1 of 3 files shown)


tembo-operator/src/snapshots/volumesnapshots.rs
```diff
@@ -271,14 +326,6 @@ fn generate_volume_snapshot(
...
             Action::requeue(tokio::time::Duration::from_secs(300))
         })?;
-    let volume_snapshot_class_name = snapshot_content
-        .spec
-        .volume_snapshot_class_name
-        .as_ref()
-        .ok_or_else(|| {
-            error!("VolumeSnapshotClass name is empty for instance {}", name);
-            Action::requeue(tokio::time::Duration::from_secs(300))
-        })?;
```
nhudson, line 281: This is removed as it's not needed in the `VolumeSnapshot` and will be populated and assumed from the `VolumeSnapshotContents`

tembo-operator/src/snapshots/volumesnapshots.rs
```diff
@@ -291,7 +338,7 @@ fn generate_volume_snapshot(
                 volume_snapshot_content_name: Some(volume_snapshot_content_name.to_string()),
                 ..VolumeSnapshotSource::default()
             },
-            volume_snapshot_class_name: Some(volume_snapshot_class_name.to_string()),
```
nhudson, line 294: This is removed as it's not needed in the VolumeSnapshot and will be populated and assumed from the VolumeSnapshotContents

tembo-operator/src/snapshots/volumesnapshots.rs
```diff
@@ -543,10 +590,6 @@ mod tests {
             result.spec.source.volume_snapshot_content_name,
             Some("test-snapshot-content".to_string())
         );
-        assert_eq!(
-            result.spec.volume_snapshot_class_name,
-            Some("test-class".to_string())
-        );
```
nhudson, line 549: Removing unneeded test

PR #596, opened 2024-02-29
adds support for extension version

shahadarsh:
* adds support to specify extension version in tembo.toml
* if the version not specified gets it from trunk registry

### DIFF
--- a/tembo-cli/src/cmd/apply.rs
+++ b/tembo-cli/src/cmd/apply.rs
@@ -26,6 +26,8 @@ use tembo::cli::tembo_config::Library;
 use tembo_stacks::apps::app::merge_app_reqs;
 use tembo_stacks::apps::app::merge_options;
 use tembo_stacks::apps::types::MergedConfigs;
+use temboclient::models::ExtensionStatus;
+use temboclient::models::Instance;
 use temboclient::{
     apis::{
         configuration::Configuration,
@@ -120,7 +122,7 @@ fn tembo_cloud_apply(
         match result {
             Ok(i) => i,
             Err(error) => {
-                tui::error(&format!("Error creating instance: {}", error));
+                tui::error(&format!("{}", error));
                 return Ok(());
             }
         }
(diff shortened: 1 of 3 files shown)


tembo-cli/src/cmd/apply.rs
```diff
@@ -561,19 +578,25 @@ fn get_postgres_config_cloud(instance_settings: &InstanceSettings) -> Vec<PgConf
...
         for (name, extension) in extensions.into_iter() {
-            let mut vec_extension_location: Vec<ExtensionInstallLocation> = vec![];
-
-            vec_extension_location.push(ExtensionInstallLocation {
-                database: Some("postgres".to_string()),
-                schema: None,
-                version: None,
-                enabled: extension.enabled,
-            });
+            let version = Runtime::new()
```
sjmiller609, line 586: the typical approach is just to make the parent function async, then use await.

tembo-cli/src/cmd/apply.rs
```diff
@@ -918,6 +933,44 @@ async fn get_loadable_libraries(
...
+    if let Some(version) = maybe_version {
+        return Ok(version);
+    }
+
+    let trunk_projects = get_trunk_projects(&name).await?;
+
+    // If more than 1 trunk_project is returned then skip getting version
+    if trunk_projects.len() > 1 {
+        return Ok("".to_string());
+    }
```
sjmiller609, line 949: this seems it should return None, not Ok("")

tembo-cli/src/cmd/apply.rs
```diff
@@ -918,6 +933,44 @@ async fn get_loadable_libraries(
...
+
+    if let Some(extensions) = trunk_project.extensions.as_ref() {
+        for trunk_extension in extensions.iter() {
+            if trunk_extension.extension_name != name {
+                continue;
+            }
+            return Ok(trunk_extension.version.clone());
+        }
+    }
+    Ok("".to_string())
```
sjmiller609, line 961: same here

tembo-cli/src/cmd/apply.rs
```diff
@@ -513,14 +530,14 @@ fn get_update_instance(instance_settings: &InstanceSettings) -> UpdateInstance {
         replicas: instance_settings.replicas,
         app_services: None,
         connection_pooler: None,
-        extensions: Some(Some(get_extensions(instance_settings.extensions.clone()))),
+        extensions: Some(Some(get_extensions(instance_settings.extensions.clone())?)),
```
sjmiller609, line 533: I think this is used on update, but it's using always the latest version of an extension found in trunk. It is not safe to assume the version should be set to latest, it should be set to the version that's already enabled in the case where the extension is already turned on.

sjmiller609 (changes requested):

tembo-cli/src/cmd/apply.rs
sjmiller609: Consider this case:

- user does not specify version
- deploy to cloud
- new version of extension published to trunk
- deploy again to same instance

sjmiller609: In the above case, it must include the previous version, not the newly published version

tembo-cli/src/cmd/apply.rs
shahadarsh: Yup, I can do that but then I will have to go all the way to the top function and add async/await for all. Do you think that's better?

sjmiller609 (approved):

PR #597, opened 2024-03-01
uses the extension version installed if new version in uploaded to trunk

shahadarsh:
Extensions can be `enabled` later so instead of skipping the extension using the version that's already installed in case it's not specified in `tembo.toml`

### DIFF
--- a/tembo-cli/src/cmd/apply.rs
+++ b/tembo-cli/src/cmd/apply.rs
@@ -634,7 +634,7 @@ fn get_extensions(
                             "Current version of extension {} installed is different than version on trunk",
                             name);
                         let ext_locations = extension_mismatch.unwrap().locations.clone();
-                        if ext_locations.len() > 0 {
+                        if !ext_locations.is_empty() {
                             if let Some(existing_version) = ext_locations[0].clone().version {
                                 version = existing_version
                             } else {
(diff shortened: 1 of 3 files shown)


tembo-cli/src/cmd/apply.rs
```diff
@@ -630,7 +630,16 @@ fn get_extensions(
                             name
                         )));
                     } else {
-                        continue;
+                        let ext_locations = extension_mismatch.unwrap().locations.clone();
+                        if ext_locations.len() > 0 {
+                            if let Some(existing_version) = ext_locations[0].clone().version {
+                                version = existing_version
+                            }
```
sjmiller609, line 637: the else for this block appears to be unhandled condition, meaning in the case where existing_version is None

tembo-cli/src/cmd/apply.rs
```diff
@@ -630,7 +630,16 @@ fn get_extensions(
                             name
                         )));
                     } else {
-                        continue;
+                        let ext_locations = extension_mismatch.unwrap().locations.clone();
```
sjmiller609, line 633: Is this unwrap definitely safe?

sjmiller609 (approved):
A couple of tips

PR #599, opened 2024-03-02
adds test for tembo cloud using minimal example

shahadarsh:


.github/workflows/cargo-test.yaml
```diff
@@ -102,3 +102,9 @@ jobs:
           set -xe
           export PROMETHEUS_URL=https://prometheus-data-1.use1.dev.plat.cdb-svc.com
           cd ${{ matrix.path }} && cargo test
+        env:
+          ORG_ID: ${{ secrets.ORG_ID }}
```
shahadarsh, line 106: Added these secrets to gh action secrets but need to make sure they are added correctly and also need to use appropriate ORG_ID & TOKEN

sjmiller609 (approved):

PR #601, opened 2024-03-04
adds stack as alias for stack_type, deny_unknown_fields

shahadarsh:
* stack as alias for stack_type
* shows below error for unknown field: 
Screenshot 2024-03-03 at 8 26 36 PM

### DIFF
--- a/tembo-cli/src/cmd/top.rs
+++ b/tembo-cli/src/cmd/top.rs
@@ -34,7 +34,7 @@ pub struct MetricsResponse {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct MetricsData {
-    pub resultType: String,
+    pub result_type: String,
     pub result: Vec<MetricResult>,
 }
 
--- a/tembo-cli/src/cmd/top.rs
+++ b/tembo-cli/src/cmd/top.rs
@@ -34,7 +34,7 @@ pub struct MetricsResponse {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct MetricsData {
-    pub result_type: String,
+    pub resultType: String,
     pub result: Vec<MetricResult>,
 }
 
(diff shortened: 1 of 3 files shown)


tembo-cli/src/cmd/top.rs
```diff
@@ -34,7 +34,7 @@ pub struct MetricsResponse {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct MetricsData {
-    pub resultType: String,
```
sjmiller609, line 37: I think this change effects parsing, so please check still works and maybe can add a test for tembo top in the test against cloud. To fix the lint issue, it’s possible to use serde macros to parse the value as something different than the variable name.

tembo-cli/src/cmd/validate.rs
```diff
@@ -49,10 +49,19 @@ pub fn execute(verbose: bool) -> Result<(), anyhow::Error> {
...
+            Ok(i) => i,
+            Err(error) => {
+                tui::error(&format!("{}", error));
+                return Ok(());
+            }
+        };
 
         // Validate the config
-        match validate_config(config, verbose) {
+        match validate_config(config.unwrap(), verbose) {
```
sjmiller609, line 64: Is this unwrap definitely safe?

sjmiller609 (approved):
Thank you!

tembo-cli/src/cmd/top.rs
shahadarsh: Oh yes.. I'm gonna revert the change.

tembo-cli/src/cmd/validate.rs
shahadarsh: Yes I there is a match here for it: https://github.com/tembo-io/tembo/pull/601/files#diff-ee5857f9bcdd882fa5af416c3a7f17c535be574357b8cf55b09945fda09407ddR55

PR #614, opened 2024-03-05
Prettify top

joshuajerin:


shahadarsh (approved):
I tested it and it looks good. The table disappears and appears back with the `--tail` option when the data gets refreshed but looks like we are good with that for now.

PR #622, opened 2024-03-06
make sure we find the closest snapshot available when pitr is requested

nhudson:
Evaluate the `recovery_target_time` if supplied and find the correct or closest `VolumeSnapshot` to restore with.

fixes: [TEM-3279](https://linear.app/tembo/issue/TEM-3279/address-finding-the-correct-snapshot-when-pitr-is-requested)

### DIFF
--- a/tembo-operator/Cargo.lock
+++ b/tembo-operator/Cargo.lock
@@ -494,7 +494,7 @@ dependencies = [
 
 [[package]]
 name = "controller"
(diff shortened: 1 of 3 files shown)


sjmiller609 (approved):

nhudson:
Apply this after we push https://github.com/tembo-io/tembo/pull/626

PR #627, opened 2024-03-07
uses alt base image based on the stack for local

shahadarsh:


sjmiller609 (approved):

sjmiller609 (approved):
it would be nice to include a test that starts on each pg version the standard stack, i.e. matrix build for local, minimal example

shahadarsh:
> it would be nice to include a test that starts on each pg version the standard stack, i.e. matrix build for local, minimal example

Yup. Working on tests in the next PR for local & cloud.

PR #636, opened 2024-03-08
make sure that the image and tag are a string

nhudson:
fixes: [TEM-3266](https://linear.app/tembo/issue/TEM-3266/investigate-where-the-controller-helm-chart-renders-an-int-instead-of)


sjmiller609:
LGTM

PR #647, opened 2024-03-08
Fix timeseries extension name

jasonmp85:
The name of the project as installed by trunk is pg_timeseries, while
"timeseries" is the extension name within PostgreSQL.


nhudson (approved):

ChuckHend (approved):

tembo-operator/src/stacks/types.rs
```diff
@@ -44,6 +44,7 @@ impl std::str::FromStr for StackType {
             "OLTP" => Ok(StackType::OLTP),
             "RAG" => Ok(StackType::RAG),
             "Standard" => Ok(StackType::Standard),
+            "Timeseries" => Ok(StackType::Timeseries),
```
ChuckHend, line 47: @nhudson this is what we missed

PR #655, opened 2024-03-12
fix logic to not update Backup during an update event

nhudson:
There is an issue where some `CoreDB`'s are being updated every 2-3 minutes and their backup schedules are changing.  This is causing backups to be taken multiple times an hour.

* Fix logic to not update the `CoreDBSpec.Backup` section on an `Event::Update`.
* Fix `generate_rand_schedule()` function to be non-async as it's not doing any async functionality
* Cleanup old commented out code.

### DIFF
--- a/conductor/src/main.rs
+++ b/conductor/src/main.rs
@@ -5,7 +5,7 @@ use conductor::extensions::extensions_still_processing;
 use conductor::monitoring::CustomMetrics;
 use conductor::{
     create_cloudformation, create_namespace, create_or_update, delete, delete_cloudformation,
-    delete_namespace, generate_rand_schedule, generate_spec, get_coredb_error_without_status,
+    delete_namespace, generate_cron_expression, generate_spec, get_coredb_error_without_status,
     get_one, get_pg_conn, lookup_role_arn, restart_coredb, types,
 };
 use controller::apis::coredb_types::{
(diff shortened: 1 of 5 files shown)


conductor/src/main.rs
```diff
@@ -649,111 +649,37 @@ async fn init_cloud_perms(
...
-        retentionPolicy: Some(String::from("30")),
-        schedule: Some(generate_rand_schedule().await),
-        s3_credentials: Some(S3Credentials {
-            inherit_from_iam_role: Some(true),
+    // If read_msg.message.event_type is Event::Update we want to use the existing backup
+    // configuration spec
+    if read_msg.message.event_type == Event::Update {
+        // Since Backup is always present, there's no need to clone it for an update.
+        // Just return here since we are not updating anything.
+    } else {
```
sjmiller609, line 657:
```suggestion
    if read_msg.message.event_type != Event::Update {
```

conductor/src/main.rs
```diff
@@ -649,111 +649,37 @@ async fn init_cloud_perms(
...
+            destinationPath: Some(format!(
+                "s3://{}/coredb/{}/org-{}-inst-{}",
+                backup_archive_bucket,
+                &read_msg.message.organization_name,
+                &read_msg.message.organization_name,
+                &read_msg.message.dbname
+            )),
+            encryption: Some(String::from("AES256")),
+            retentionPolicy: Some(String::from("30")),
+            schedule: Some(generate_rand_schedule()),
```
sjmiller609, line 669: It seems to me like this would reset the schedule to null in the case of an Update, perhaps? It would do an apply of the CoreDB resource type without adding in these configurations?

nhudson: the update is just updating the `CoreDBSpec`.  By not updating the `spec.backup` section during an update it should reuse the same values since we are not overriding or updating anything.  Or that was my thought anyway.

sjmiller609 (approved):

PR #658, opened 2024-03-13
Generate tembo.toml file locally

joshuajerin:


tembo-cli/src/cmd/init.rs
```diff
@@ -42,11 +44,21 @@ pub fn execute() -> Result<(), anyhow::Error> {
         }
     }
 
-    let filename = "tembo.toml";
-    let filepath =
-        "https://raw.githubusercontent.com/tembo-io/tembo/main/tembo-cli/examples/single-instance/tembo.toml";
+    let cargo_manifest_dir =
+        env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR env var not set");
```
sjmiller609, line 48: I think this logic only works because you are running from source, but would not work after build + distributing the binary

joshuajerin: You're right. I changed the code accordingly

sjmiller609 (approved):

tembo-cli/src/cli/file_utils.rs
```diff
@@ -53,36 +52,13 @@ impl FileUtils {
         Ok(())
     }
 
-    pub fn download_file(
-        file_path: &str,
-        download_location: &str,
-        recreate: bool,
-    ) -> std::io::Result<()> {
-        let path = Path::new(&download_location);
-        if !recreate && path.exists() {
```
shahadarsh, line 62: @joshuajerin The logic to skip generating the `tembo.toml` file if it exists should stay otherwise init will replace existing tembo.toml file which we don't want.

PR #660, opened 2024-03-14
Flags for tembo login

joshuajerin:


tembo-cli/src/cmd/login.rs
```diff
@@ -168,3 +211,92 @@ pub fn update_access_token(
 
     Ok(())
 }
+
+pub fn update_context(org_id: &str, profile_name: &str) -> Result<(), anyhow::Error> {
```
sjmiller609, line 215: I think this way of writing the file would replace the entire content of the file, for example re-ordering the file, removing comments, etc. Maybe instead we rended the new part into a block string in memory, then append it to the end of the file.

tembo-cli/src/cmd/login.rs
```diff
@@ -168,3 +211,92 @@ pub fn update_access_token(
...
+    };
+    data.environment.push(new_env);
+
+    let modified_contents = toml::to_string(&data)?;
+    fs::write(&context_file_path, modified_contents)?;
+
+    Ok(())
+}
+
+pub fn update_or_create_profile(
```
sjmiller609, line 236: same idea here regarding not replacing the whole content of the file, rather appending to a file.

tembo-cli/src/cmd/login.rs
```diff
@@ -11,28 +12,60 @@ use serde::Deserialize;
...
-        .as_ref()
-        .ok_or_else(|| anyhow!("Cannot log in to the local context, please select a tembo-cloud context before logging in"))?;
-    let login_url = url(profile)?;
-    let rt = tokio::runtime::Runtime::new().expect("Failed to create a runtime");
 
-    rt.block_on(handle_tokio(login_url))?;
+    match (&login_cmd.organization_id, &login_cmd.profile) {
+        (Some(_), None) | (None, Some(_)) => {
+            return Err(anyhow!(
+                "Both 'organization_id' and 'profile' must be specified."
```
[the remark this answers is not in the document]
joshuajerin: Yes. We return the error there when either the --organization_id or --profile flags aren't specified

tembo-cli/src/cmd/login.rs
joshuajerin: I was able to test the code and it doesn't look like it replaces nor re-orders the file but appends to the end of the file with the new profile.

tembo-cli/src/cmd/login.rs
joshuajerin: Same here

tembo-cli/src/cmd/init.rs
```diff
@@ -1,16 +1,32 @@
...
+        CONTEXT_EXAMPLE_TEXT
+    } else {
+        CONTEXT_DEFAULT_TEXT
+    };
+
+    let credentials_text = if is_test_env {
+        CREDENTIALS_EXAMPLE_TEXT
+    } else {
+        CREDENTIALS_DEFAULT_TEXT
+    };
```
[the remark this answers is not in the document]
joshuajerin: Could you tell me how I could approach this problem for the tests since it wouldn't run without any profiles set up? I don't think I could run login because it opens up mahout which won't work in github actions

tembo-cli/src/cmd/login.rs
```diff
@@ -171,3 +211,91 @@ pub fn update_access_token(
...
+    let mut contents = fs::read_to_string(&context_file_path)?;
+    let mut data: Context = toml::from_str(&contents)?;
+
+    if let Some(env) = data.environment.iter_mut().find(|p| p.set == Some(true)) {
+        env.set = Some(false);
+    }
+
+    match data.environment.iter().any(|p| p.name == profile_name) {
+        true => {
+            error(&format!("Context Environment already exists. Either try creating a new name or set it to that."));
```
joshuajerin, line 226: @sjmiller609 Could you confirm if this error message is good when user tries to create a duplicate profile? Thank you :)

sjmiller609 (approved):

PR #661, opened 2024-03-14
run backup before setting up replica

nhudson:
Since enabling snapshots with CNPG, we need a way to bring a potential new replica inline with the primary instance.  Currently CNPG states that you must take a backup before bringing a new replica online.

We have opened an issue with CNPG in hopes to build a better solution to this problem. https://github.com/cloudnative-pg/cloudnative-pg/issues/4125

Currently the logic works like this
if increasing replicas (1->2)
* If there was a backup taken in the last 45 minutes, continue bringing the new replica online.
  * Otherwise, take a backup
* If there is a backup in `pending`  or `running`
  * requeue. 
* Continue to rest of code that increases the replicas

fixes: [PLAT-117](https://linear.app/tembo/issue/PLAT-117/bug-ha-replica-fails-to-start)


tembo-operator/tests/integration_tests.rs
```diff
@@ -4166,7 +4166,8 @@ CREATE EVENT TRIGGER pgrst_watch
         // Wait for Postgres to restart
         {
             let started = Utc::now();
-            let max_wait_time = chrono::Duration::seconds(TIMEOUT_SECONDS_POD_READY as _);
+            let max_wait_time =
+                chrono::TimeDelta::try_seconds(TIMEOUT_SECONDS_POD_READY as _).unwrap();
```
nhudson, line 4170: Updating deprecated method https://docs.rs/chrono/latest/chrono/type.Duration.html#method.seconds

tembo-operator/tests/integration_tests.rs
```diff
@@ -4220,7 +4221,7 @@ CREATE EVENT TRIGGER pgrst_watch
     async fn functional_test_status_configs() {
         async fn runtime_cfg(coredbs: &Api<CoreDB>, name: &str) -> Option<Vec<PgConfig>> {
             let started_waiting = Utc::now();
-            let max_wait_time = chrono::Duration::seconds(45);
+            let max_wait_time = chrono::TimeDelta::try_seconds(45).unwrap();
```
nhudson, line 4224: Updating deprecated method https://docs.rs/chrono/latest/chrono/type.Duration.html#method.seconds

sjmiller609 (approved):

ianstanton (approved):

PR #671, opened 2024-03-21
fix: databases recreated when delete during provisioning

joshuajerin:


.github/workflows/cargo-test.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
         run: |
           set -xe
           sudo apt-get update
+          cargo install sqlx-cli
```
sjmiller609, line 47:
```suggestion
```

conductor/migrations/20240321181458_create_deleted_instances_table.up.sql
```diff
@@ -0,0 +1,8 @@
+-- Up migration
+CREATE TABLE deleted_instances (
+    namespace VARCHAR(255) NOT NULL,
+    deleted_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (namespace)
+);
+
+CREATE UNIQUE INDEX idx_deleted_instances_namespace ON deleted_instances(namespace);
```
ChuckHend, line 8: Is this writing to the queue instance or the control-plane instance?

If this is the queue instance, then we should make sure that the queue instance is dedicated to the one cluster. Up to this point the control-plane's postgres instance has been the sole location for any state related to a data-plane.

If this is the control-plane's DB, then note that this is the first introduction of the pattern of data-plane writing directly back to the control-plane database.

cc @sjmiller609 , @ianstanton

sjmiller609:
needs SQLX_OFFLINE in the dockerfile

.github/workflows/deploy.yml
```diff
@@ -44,12 +44,41 @@ jobs:
       - self-hosted
       - dind
       - large-8x8
+    env:
```
sjmiller609, line 47: I think this file should not have any changes

conductor/Dockerfile
```diff
@@ -23,3 +25,4 @@ RUN set -eux; \
     apt-get clean; \
     rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 COPY --from=builder /usr/local/cargo/bin/* /usr/local/bin/
+COPY --from=builder /build/migrations /migrations
```
sjmiller609, line 28: I think this line is not necessary

conductor/migrations/20240321181458_create_deleted_instances_table.up.sql
sjmiller609: We discussed in Slack. It is the queue DB, not the control plane DB, and I think we are both OK with this approach to save a table of deleted instances in the queue db.

sjmiller609 (changes requested):

conductor/src/main.rs
```diff
@@ -96,6 +113,43 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
...
+            .await
+            {
+                Ok(Some(_)) => true,
+                Ok(None) => false,
+                Err(e) => {
+                    error!("Database query error: {}", e);
+                    continue;
+                }
+            };
+            println!("{}", namespace_exists);
```
sjmiller609, line 137: remove this print statement

conductor/src/main.rs
```diff
@@ -96,6 +113,43 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
...
         let org_id = read_msg.message.org_id.clone();
         let instance_id = read_msg.message.inst_id.clone();
+        let namespace = format!(
+            "org-{}-inst-{}",
+            read_msg.message.organization_name, read_msg.message.dbname
+        );
+        info!("{}: Using namespace {}", read_msg.msg_id, &namespace);
+
+        if read_msg.message.event_type != Event::Delete {
+            let namespace_exists = match sqlx::query!(
```
sjmiller609, line 123: rename variable to "namespace_already_deleted"

sjmiller609 (approved):

PR #676, opened 2024-03-25
fix: don't assign new service name to legacy tcp routes

sjmiller609:


tembo-operator/src/ingress.rs
```diff
@@ -274,7 +274,7 @@ pub async fn reconcile_postgres_ing_route_tcp(
     let mut present_ing_route_tcp_names_list: Vec<String> = vec![];
 
     // Check for all existing IngressRouteTCPs in this namespace
-    // Filter out any that are not for this DB or do not have the correct prefix
+    // Filter out any that are not for this DB, do not have the correct prefix, or do not have matching service name
```
sjmiller609, line 277: This was originally programmed like this to aid in the migration from a different service name and domain name to new service name and domain name, but now we only need to protect from domain name swapping. The point of this is so that if we add different domain names to the data plane, it doesn't ever remove one of the domains from the ingressroutetcp, which would break existing connections, for example instances that are still connected via `...coredb.io` domain

sjmiller609: but I think we don't need this any more, so I just simplified the whole block

tembo-operator/src/app_service/manager.rs
```diff
@@ -602,7 +602,7 @@ async fn apply_resources(resources: Vec<AppServiceResources>, client: &Client, n
 fn generate_appsvc_annotations(cdb: &CoreDB) -> BTreeMap<String, String> {
     cdb.metadata.annotations.as_ref().map_or_else(
         || {
-            error!(
+            debug!(
```
sjmiller609, line 605: normal in case of local dev

nhudson (approved):

PR #696, opened 2024-04-04
add kubernetes 429 error requeue with jitter

nhudson:
Add error handling in the client to requeue differently when a 429 is returned back from the server.

* Added two new tests to mock the Kubernetes API to test a 429 return code and a non 429 return code.

Fixes: [PLAT-497](https://linear.app/tembo/issue/PLAT-497/handle-kube-api-429-with-requeue)


sjmiller609 (approved):
This is very elegant

PR #697, opened 2024-04-04
Remove `password` from `ConnectionInfo`

vrmiguel:


conductor/src/types.rs
```diff
@@ -59,7 +59,6 @@ pub struct ConnectionInfo {
     pub pooler_host: Option<String>,
     pub port: u16,
     pub user: String,
-    pub password: String,
     pub app_user: String,
     pub app_password: String,
```
sjmiller609, line 63: what about app password?

conductor/src/lib.rs
```diff
@@ -303,7 +303,6 @@ pub async fn get_pg_conn(
         pooler_host,
         port: 5432,
         user: postgres_user,
-        password: postgres_pw,
         app_user,
         app_password: app_pw,
```
sjmiller609, line 307: same

conductor/src/types.rs
sjmiller609: I thought there is also readonly user, but I'm not sure where that info is

sjmiller609 (approved):
This looks right to me, but I think it may break on deserialization on the control plane end, so you have to carefully coordinate right correct order of merges for this one.

Good to go when you believe it's the right order of merging, i.e. issues handled control plane side, and you also make sure app and readonly users' passwords not getting sent to control plane either

PR #701, opened 2024-04-09
Fix: Remove Postgres password from path

joshuajerin:


dataplane-webserver/src/routes/secrets.rs
```diff
@@ -251,20 +256,23 @@ fn is_valid_id(s: &str) -> bool {
...
+    request_body = PasswordString,
     responses(
         (status = 200,
             description = "Password successfully changed."),
         (status = 403,
             description = "Not authorized for query"),
     )
 )]
-#[patch("/secrets/{secret_name}/passwords/{password}")]
+#[patch("/secrets/{secret_name}/passwords")]
```
sjmiller609, line 267: Can you please change the path to
```
/api/v1/orgs/{org_id}/instances/{instance_id}/secrets/{secret_name}
```

sjmiller609: Input validation contains at least:

admin role in the org they are trying to modify

only allow the secret names "app-role", "readonly-role", and "superuser-role"

validate the password is at least 16 characters long

sjmiller609 (approved):
Only thing is making sure only the 3 secret names are allowed "app-role", "readonly-role", and "superuser-role"

sjmiller609 (approved):

PR #716, opened 2024-04-17
Send metrics from dataplane to control-plane's queue

vrmiguel:


sjmiller609 (approved):

sjmiller609 (approved):
This will fix the conductor error about traefik https://github.com/tembo-io/tembo/pull/717/commits/8cf796d706dc791c98bcde50d9e8954d86ff57de

and I think in the docker build, I think it just needs a COPY command for the metrics file

PR #717, opened 2024-04-17
Support node/pod affinity, tolerations and topologySpreadConstraints that CNPG supports

nhudson:
The PR will add a new API field called `affinity_configuration` that is a type `ClusterAffinity` and `topology_spread_constraints` that is type `ClusterTopologySpreadConstraints` from CNPG.  This will allow us to directly support all of the node/pod affinity and toleration types that CNPG supports.  

The concept is that the control-plane will set the toleration/affinity when setting up the new instance, which we will just pass on to CNPG using their [spec](https://cloudnative-pg.io/documentation/1.22/cloudnative-pg.v1/#postgresql-cnpg-io-v1-AffinityConfiguration).

```yaml
apiVersion: coredb.io/v1alpha1
kind: CoreDB
metadata:
  name: test-db-restore
spec:
  affinityConfiguration:
    podAntiAffinityType: preferred
    topologyKey: topology.kubernetes.io/zone
    nodeSelector: tembo.io/provisioner: spot
    tolerations:
      - key: "tembo.io/spot"
        operator: "Equal"
        value: "true"
        effect: "NoSchedule"
  topologySpreadConstraints:
    - maxSkew: <integer>
      minDomains: <integer> # optional
      topologyKey: <string>
      whenUnsatisfiable: <string>
      labelSelector: <object>
 ```

This change will move the default setting of `affinity` of CNPG from `cnpg_cluster_from_cdb` function into the `CoreDB` API with our current defaults

```yaml
    podAntiAffinityType: preferred
    topologyKey: topology.kubernetes.io/zone
```

This will also do the same for the `Pooler` API and also the `AppService` `Deployments`.

The `PlacementConfig` does a lot of work specifically around converting the types.  So we take in the `ClusterAffinity` and `ClusterTopologySpreadConstraints` they get converted into native Kubernetes API's and placed in `PlacementConfig`.  The `PlacementConfig` has several methods that can convert the types to use with the `Pooler` or we can use the types directly in a normal Kubernetes `Pod` or `Deployment` if we wish.

fixes: [PLAT-431](https://linear.app/tembo/issue/PLAT-431/add-to-operator-node-placement-features)


ianstanton (approved):
LGTM once everything is passing :tada:

sjmiller609 (approved):

sjmiller609 (approved):

PR #719, opened 2024-04-18
SPAE-141 add pg_tier extn to dwh stack

shhnwz:
SPAE-141 add pg_tier extn to dwh stack

### DIFF
--- a/tembo-operator/Cargo.toml
+++ b/tembo-operator/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.44.2"
+version = "0.44.1"
 edition = "2021"
 default-run = "controller"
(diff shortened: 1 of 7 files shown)


tembo-operator/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.44.1"
+version = "0.44.2"
```
ChuckHend, line 4: No need to bump the operator version on this. Just bump version on tembo-stacks/Cargo.toml

ChuckHend (changes requested):
Looks good, just undo the change to the operator's Cargo.toml

tembo-operator/Cargo.toml
shhnwz: reverted the increment. have a look

PR #720, opened 2024-04-18
Add support for CNPG Hibernation functionality

bonesmoses:
In order to cover PLAT-467, we needed to provide functionality to toggle the CNPG hibernation notation. This branch should accomplish that. Functional test included.


tembo-operator/tests/integration_tests.rs
```diff
@@ -65,6 +65,112 @@ mod test {
...
+    /// 
+    /// Most if not all tests use all of the fields here, or refer to them
+    /// in some manner. This struct helps combine everything together.
+    struct TestClass {
+        name: String,
+        namespace: String,
+        client: Client,
+        state: State,
+        context: Arc<Context>,
+        pods: Api<Pod>,
```
nhudson, line 78: I really like this concept and it's maybe something we can go back and adjust for all tests.  

The only issue is that these three objects in the struct are not being used.  I suggest to comment them out for now until they are used.

tembo-operator/src/controller.rs
```diff
@@ -378,23 +383,16 @@ impl CoreDB {
...
-                    trunk_installs: self.status.as_ref().and_then(|f| f.trunk_installs.clone()),
-                    resources: Some(self.spec.resources.clone()),
-                    runtime_config: Some(current_config_values),
-                    first_recoverability_time: self
-                        .status
-                        .as_ref()
-                        .and_then(|f| f.first_recoverability_time),
-                    pg_postmaster_start_time: None,
-                    last_fully_reconciled_at: None,
-                }
```
nhudson, line 397: I am not really sure why you're removing these.

tembo-operator/src/controller.rs
```diff
@@ -378,23 +383,16 @@ impl CoreDB {
...
+                let mut status = self.status.clone().unwrap_or_default();
```
nhudson, line 391: I would not use `unwrap_or_default` anywhere.  If for some reason if status is empty this will panic, or potentially panic.  A possibly solution might be this, but I would make sure this works before applying it.

```rust
let status = match self.status.clone() {
    Some(s) => s,
    None => return Err(Action::requeue(Duration::from_secs(300)))
};
```

nhudson (changes requested):

tembo-operator/src/controller.rs
```diff
@@ -345,11 +345,16 @@ impl CoreDB {
...
-        let pg_postmaster_start_time = is_not_restarting(self, ctx.clone(), "postgres").await?;
+        // This block of code is currently fairly contentious. What status
+        // elements we should retain in the event the cluster is in a stop
+        // state are currenly unknown and under review. This appears to work
+        // for now, but care should be taken given this status.
 
         let mut new_status = match self.spec.stop {
             false => {
+                // Check if Postgres is already running
+                let pg_postmaster_start_time = is_not_restarting(self, ctx.clone(), "postgres").await?;
```
nhudson, line 356: I believe this line is causing the fmt/lint CI job to fail.  Running `cargo fmt` should clear it up.

nhudson (changes requested):

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1025,8 +1060,25 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
     // Check if the CoreDB status is running: false, return requeue
     if let Some(status) = current_status {
         if !status.running {
-            info!("CoreDB status.running is false, requeuing 10 seconds");
-            return Err(Action::requeue(Duration::from_secs(10)));
+            // Status is not running, we should check if the cluster is ready, if not requeue
+            match check_cluster_status(cdb, &ctx).await {
+                Ok(true) => {
+                    info!("CoreDB status.running is false, but cluster is ready, update CoreDB status.running to true for instnace: {}", &name);
```
sjmiller609, line 1066:
```suggestion
                    info!("CoreDB status.running is false, but cluster is ready, update CoreDB status.running to true for instance: {}", &name);
```

tembo-operator/src/controller.rs
sjmiller609: unwrap or default doesn't panic when unwrap fails, instead it returns the default of the desired type (CoreDBStatus)

tembo-operator/src/controller.rs
sjmiller609: This part I think makes sense. Basically it is leave the status as-is on the K8s resource when you stop the instance, except it's clearing out pg_postmaster_start_time, setting last_fully_reconciled_at, and setting running to false

tembo-operator/tests/integration_tests.rs
```diff
@@ -65,6 +65,112 @@ mod test {
     const TIMEOUT_SECONDS_POD_DELETED: u64 = 300;
     const TIMEOUT_SECONDS_COREDB_DELETED: u64 = 300;
 
+    /// Struct to contain many commonly used test resources
+    /// 
+    /// Most if not all tests use all of the fields here, or refer to them
+    /// in some manner. This struct helps combine everything together.
+    struct TestClass {
```
sjmiller609, line 72: I also like this setup, but I would just change the naming of this, perhaps to "TestFixture". Also I am being perhaps picky, but I would just remove all mentions of the word "class", as we don't literally have classes is Rust. The word "Fixture" I think is the generic word used for what this is doing.

sjmiller609 (approved):
Very nice, cleanly done. I like how the new code is clearly separated from existing. It seems to have bare-minimum touching of the existing code, which is perfect, i.e. the minimum number of lines modified to get the job done, and new code is in new functions. Beautiful!

tembo-operator/src/controller.rs
bonesmoses: Steve recommended that, and I agreed at the time, but again, I don't know what's actually happening there.

tembo-operator/tests/integration_tests.rs
bonesmoses: Yeah... I was kinda hoping there was a way to just suppress those specific warnings because there's full intent to use those.

bonesmoses: Now that I reexamine, I think a better solution than commenting these out would be to provide a get interface rather than reaching deep into the struct. These elements in particular are often used directly in the other tests, so need a better assignment paradigm. That would solve both problems at once. I'll patch that in.

bonesmoses: On second thought, that just creates more functions that will remain "unused" and add back into the code duping issue. I'll just convert the basic-cnpg test which uses these.

tembo-operator/tests/integration_tests.rs
bonesmoses: I've opted for `TestCore` just to get rid of the nomenclature confusion. I couldn't find any consensus on the word "Fixture" in the context of Rust, but perhaps my keywords were flawed if this is a more generic-concept.

nhudson (approved):

bonesmoses:
Is it possible to merge while ignoring the functional tests? I've run them 4 times now, and different ones fail ever time. Additionally, we should track down the race conditions that keep causing this.

sjmiller609 (changes requested):
Needs to also stop app services by setting replicas to zero

sjmiller609 (approved):

PR #730, opened 2024-04-22
Simplify stop code

sjmiller609:
This is what it looks like, as compared to main https://github.com/tembo-io/tembo/pull/731/files


tembo-operator/tests/integration_tests.rs
```diff
@@ -1292,7 +1267,18 @@ mod test {
         let params = PatchParams::apply("coredb-integration-test");
         let patch = Patch::Apply(&coredb_json);
         let _ = coredbs.patch(name, &params, &patch).await.unwrap();
-        thread::sleep(Duration::from_millis(10000));
```
sjmiller609, line 1295: This code assumes it will finish in 10 seconds, which is not the case in a busy cluster

tembo-operator/tests/integration_tests.rs
```diff
@@ -3060,9 +3056,9 @@ mod test {
         wait_until_psql_contains(
             context.clone(),
             coredb_resource.clone(),
-            "select extname from pg_catalog.pg_extension;".to_string(),
```
sjmiller609, line 3063: This query was previously doing nothing

sjmiller609:
@bonesmoses you may please accept this PR into your branch if you are happy with it

bonesmoses (commented):
I like the extra module file better. Putting the hibernation reconcile at the very beginning of the process right after `coredbs` is initialized is a better approach too.

PR #736, opened 2024-04-23
adds step to bump homebrew formula automatically for CLI

shahadarsh:
* Triggers on an update to Cargo.toml file for CLI 
* Does a GitHub release
* Automatically updates Homebrew repo with the new version


sjmiller609 (approved):
Thank you!

PR #743, opened 2024-04-24
Stack metrics

ChuckHend:
- Adds metrics to the vectordb stack
- discover databases to scrape metrics from (mq and vectordb stacks)
- installs stack's extensions into the `template1` database so that any additional databases created in the Stack instance will also have the extensions automatically installed in the new database


tembo-operator/src/extensions/database_queries.rs
```diff
@@ -33,7 +33,7 @@ pub fn check_input(input: &str) -> bool {
 pub const LIST_SHARED_PRELOAD_LIBRARIES_QUERY: &str = r#"SHOW shared_preload_libraries;"#;
 
 pub const LIST_DATABASES_QUERY: &str =
-    r#"SELECT datname FROM pg_database WHERE datistemplate = false;"#;
+    r#"SELECT datname FROM pg_database WHERE datname != 'template0';"#;
```
ChuckHend, line 36: only exclude `template0`. That database is not configured to accept any connections. `template1` does allow it though, and `template1` is where we will install Stack extensions into.

tembo-stacks/src/stacks/specs/vectordb.yaml
```diff
@@ -123,3 +129,17 @@ postgres_config:
...
+      metrics:
+        - total_jobs:
+            usage: GAUGE
+            description: Number of vectorize jobs
+        - datname:
+            usage: "LABEL"
+            description: "Name of current database"
+      target_databases:
+        - "*"
+        - "postgres"
```
sjmiller609, line 145: is "postgres" redundant with "*"?

sjmiller609 (approved):

tembo-stacks/src/stacks/specs/vectordb.yaml
ChuckHend: Yes I think so, but the [cnpg example](https://cloudnative-pg.io/documentation/1.16/monitoring/) also lists a database when using the wildcard `*`

nhudson (approved):

PR #748, opened 2024-04-25
Bump timeseries for release

jasonmp85:
Getting compression out there.

### DIFF
--- a/tembo-stacks/Cargo.lock
+++ b/tembo-stacks/Cargo.lock
@@ -2294,7 +2294,7 @@ dependencies = [
 
 [[package]]
 name = "tembo-stacks"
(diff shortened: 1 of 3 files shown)


ChuckHend (approved):
conductor test probably ran because of the change to `conductor/Cargo.lock`. We could probably remove that change?

nhudson (approved):

jasonmp85:
> conductor test probably ran because of the change to `conductor/Cargo.lock`. We could probably remove that change?

Yeah, VS Code or whatever modified the file and I noticed in some others' commits that they included it so I put it in to save others the time, but it seems to have failed at any rate. I'll back it out.

PR #754, opened 2024-04-29
Don't use instance name or org name to determine namespace

sjmiller609:
Use the namespace field provided by control plane instead of using org name and instance name to determine what namespace to deploy into.

### DIFF
--- a/conductor/src/lib.rs
+++ b/conductor/src/lib.rs
@@ -382,9 +382,9 @@ pub async fn restart_coredb(
 // This will create an IAM role for the database to use to access the backup archive bucket
 pub async fn create_cloudformation(
     aws_region: String,
-    backup_archive_bucket: String,
-    org_name: &str,
-    db_name: &str,
+    bucket_name: String,
+    org_id: &str,
+    namespace: &str,
     cf_template_bucket: &str,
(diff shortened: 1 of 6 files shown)


conductor/src/lib.rs
```diff
@@ -395,26 +395,20 @@ pub async fn create_cloudformation(
...
-        String::from(db_name),
-        // AWS IAM Role Name to create
-        String::from(&iam_role_name),
-        // The AWS S3 Bucket where the CF Template is placed
-        String::from(cf_template_bucket),
-        // The Kubernetes Namespace where the database is deployed
+    let namespace = namespace.to_string();
+    let stack_name = format!("{}-cf", namespace);
+    let iam_role_name = format!("{}-iam", namespace);
+    let service_account_name = format!("{}-sa", namespace);
```
sjmiller609, line 401: These values should be the same on existing instances

ChuckHend (approved):

sjmiller609:
I noticed I need to consider the case of permissions for read-within-org for instance restore, since the org_id path prefix will not exist for existing instances, rather it will be using org name

sjmiller609:
duplicate of https://github.com/tembo-io/tembo/pull/773

PR #758, opened 2024-05-04
expose wal archiver timestamp

nhudson:
Expose the timestamp of the WAL archive from CNPG into `CoreDBStatus`.  This will allow us to better find a PITR timestamp and ensure that there is an archive when a user requests a specific time to restore to.

* includes some code cleanup clippy introduced in 1.78.0.

fixes: [PLAT-674](https://linear.app/tembo/issue/PLAT-674/find-work-around-for-restore-issue)


ChuckHend (approved):

sjmiller609 (approved):
Can you please double check that this doesn't cause repeated reconciliations, then looks good.

PR #761, opened 2024-05-06
update stack specs with equivalent constraints

ChuckHend:
ML stack is the only one that has any constraints on CPU or Mem.

Depends on #760


jasonmp85 (approved):
I like this diff.

sjmiller609 (approved):
I think CLI might use these for locally deploying a stack

ChuckHend:
> I think CLI might use these for locally deploying a stack

Took a peak and couldn't find anywhere that they are used. However, it is defined in an example [here](https://github.com/tembo-io/tembo/blob/7469149b3d88d5a7e5eda54daff9e3ba614109f1/tembo-cli/examples/stack-file/sample-stack-file.yaml#L10)

PR #763, opened 2024-05-07
remove Stacks from tembo-controller

ChuckHend:
Removes all Stack's objects from the tembo-controller. These are added to the tembo-stacks crate in https://github.com/tembo-io/tembo/pull/764


tembo-operator/src/defaults.rs
```diff
@@ -12,8 +6,25 @@ use crate::{
...
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema, PartialEq, ToSchema)]
+pub struct ImagePerPgVersion {
+    #[serde(rename = "14")]
+    pub pg14: String,
+    #[serde(rename = "15")]
+    pub pg15: String,
+    #[serde(rename = "16")]
+    pub pg16: String,
+}
```
ChuckHend, line 27: this moved here from `src/stacks/types.rs`. I wasn't sure a better place for it to go.

sjmiller609 (approved):
Nice :D

PR #764, opened 2024-05-07
move Stacks objects to tembo-stacks

ChuckHend:
Moves Stack's objects from tembo-controller to tembo-stacks.

This is in separate PR due to dependency on `ImagePerPgVersion`, which was previously in tembo-operator/src/stacks, and that directly no longer exists. So, we need to publish that crate first, then update the tembo-stacks crate.


tembo-stacks/src/stacks/config_engines.rs
```diff
@@ -0,0 +1,513 @@
+use anyhow::Result;
```
ChuckHend, line 1: formerly tembo-operator/src/stacks/config_engines.rs

sjmiller609 (approved):

tembo-stacks/src/stacks/config_engines.rs
jasonmp85: Shouldn't it be a move/rename, then? Doing it this way will obliterate any history this file has.

ChuckHend: Depends.

PR #773, opened 2024-05-08
Use s3 read and write paths from control plane

sjmiller609:
WIP, opening so I can see what needs to change on control plane side to handle the new version of cloudformation template

### DIFF
--- a/conductor/src/main.rs
+++ b/conductor/src/main.rs
@@ -118,10 +118,7 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
 
         let org_id = &read_msg.message.org_id;
         let instance_id = &read_msg.message.inst_id;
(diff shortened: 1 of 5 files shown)


conductor/src/main.rs
```diff
@@ -736,18 +724,16 @@ async fn init_cloud_perms(
...
 
-    let instance_name_slug = format!(
-        "org-{}-inst-{}",
-        &read_msg.message.organization_name, &read_msg.message.dbname
-    );
+    let write_path = read_msg
+        .message
+        .backups_write_path
+        .clone()
+        .unwrap_or(format!("v2/{}", read_msg.message.namespace));
```
sjmiller609, line 731: de-dupe this format of default, also in another location

nhudson (approved):
:rocket: LGTM!

PR #782, opened 2024-05-13
CLI: use namespace instead of org name

sjmiller609:


vrmiguel (approved):

sjmiller609:
just need to run pipeline on main twice to publish temboclient 1.0.2 then tembo CLI, that will resolve this CI error

PR #790, opened 2024-05-17
Attempt to fix pg_cron install

jasonmp85:
I'm getting a weird error on stack creation that I can't explain and the only difference between the time-series stack and others is that I didn't explicitly list all extensions in this section, instead re- lying on `CASCADE`. It seems like that may not work like I thought.


ChuckHend (approved):

ChuckHend:
Trunk has not had a pg_cron update for 4 months: https://pgt.dev/extensions/pg_cron

jasonmp85:
> Trunk has not had a pg_cron update for 4 months: https://pgt.dev/extensions/pg_cron

Yes, but several stacks are using 1.5.2, not 1.6, and it seems this might be related to 16.1? I'm not sure.

PR #796, opened 2024-05-21
Enabled pg_tier for olap, mq stacks

shhnwz:
Closes: SPAE-193

### DIFF
--- a/tembo-stacks/src/stacks/specs/oltp.yaml
+++ b/tembo-stacks/src/stacks/specs/oltp.yaml
@@ -38,9 +38,23 @@ postgres_config:
 trunk_installs:
   - name: pg_stat_statements
     version: 1.10.0
+  - name: parquet_s3_fdw
+    version: 1.1.0
+  - name: pg_tier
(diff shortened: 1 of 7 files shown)


tembo-stacks/src/stacks/specs/oltp.yaml
```diff
@@ -38,9 +38,23 @@ postgres_config:
 trunk_installs:
   - name: pg_stat_statements
     version: 1.10.0
+  - name: parquet_s3_fdw
```
ChuckHend, line 41: i didnt think we wanted this in the OLTP stack by default

ChuckHend (changes requested):
Can you change this from oltp to olap stack?

tembo-stacks/src/stacks/specs/oltp.yaml
shhnwz: Ok, I will remove pg_tier from OLTP stack.

ChuckHend (approved):

PR #802, opened 2024-05-23
Add Region to CF Bucket Endpoint URL

ianstanton:
Buckets in AWS GovCloud (and possibly other regions) require the region in the bucket endpoint URL. Add this to CloudFormation template URL. 

If region is `us-east-1`, fall back to original URL format to avoid extra changes cloud.


ianstanton:
This would require us to update conductor's IAM policy, so I'll have it fall back to the original format if region is `us-east-1`

nhudson (approved):

PR #804, opened 2024-05-23
error handling for inf gateway

ChuckHend:
Also confirming this change roll out to the correct environments.


jasonmp85 (approved):
A bit repetitive, but gets the job done.

PR #809, opened 2024-05-28
add podmonitor to coredb spec

ChuckHend:
adds metrics to AppService spec:

```yaml
app_services:
  metrics:
    path: /metrics
    port: 3000
```

When this is present in the CoreDB Spec, Tembo Operator deploys a PodMonitor for the specified AppService, which will result in scraping metrics from the specified port and path.

The PodMonitor is owned by the `Service` associated with the AppService, so when the AppService is deleted, so is the Service and the PodMonitor.


sjmiller609 (approved):

nhudson (approved):
You might want to bump the controller version as well.

PR #810, opened 2024-05-28
return the created or restored event before waiting, force wait in cp

nhudson:


conductor/src/main.rs
```diff
@@ -368,29 +368,29 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
...
-                        )
-                        .await?;
-                    metrics.conductor_requeues.add(
-                        &opentelemetry::Context::current(),
-                        1,
-                        &[KeyValue::new("queue_duration", "short")],
-                    );
-                    continue;
-                }
+                // let extension_still_processing = match &result {
```
sjmiller609, line 371: Can you please just delete instead of commenting out

nhudson: oops yeah, thats my bad.

sjmiller609 (approved):

PR #816, opened 2024-05-29
Tembo AI Helm chart

sjmiller609:


charts/tembo-ai/templates/inference-gateway/hpa.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.inferenceGateway.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
```
sjmiller609, line 2:
```suggestion
apiVersion: autoscaling/v2
```

.github/workflows/helm-test.yaml
```diff
@@ -50,17 +50,24 @@ jobs:
           cd code
           git fetch origin main:main
 
+      - name: Check chart version
+        run: |
```
nhudson, line 54: This is here twice it looks like.

charts/tembo-ai/values.yaml
```diff
@@ -0,0 +1,144 @@
...
+    match: ~
+  serviceAccount:
+    create: true
+    annotations: {}
+  service:
+    port: 8080
+  args: []
+  command:
+    - "/usr/local/bin/gateway"
+  env: []
```
ChuckHend, line 55: can we assign env vars from secrets here? `DATABASE_URL` will need to be set on the gateway server.

nhudson (approved):
Only see one possible change.  Otherwise LGTM!

charts/tembo-ai/values.yaml
```diff
@@ -0,0 +1,144 @@
...
+    portName: metrics
+    containerPort: 8081
+  serviceAccount:
+    create: true
+    annotations: {}
+  service:
+    port: 8000
+  args: []
+  command: []
+  env: []
```
ChuckHend, line 118: Same question above re: environment variables. We will need to set `HUGGING_FACE_HUB_TOKEN` as an env var from secret.

charts/tembo-ai/values.yaml
```diff
@@ -0,0 +1,144 @@
...
+    value: "true"
+    effect: "NoSchedule"
+  affinity: {}
+  podAnnotations: {}
+  podSecurityContext: {}
+  persistence:
+    enabled: true
+    size: 100Gi
+    storageClass: ""
+    accessMode: ReadWriteOnce
```
ChuckHend, line 143: Any problems if we change this to `ReadOnlyMany` in the future?

ChuckHend (approved):

ianstanton (approved):

charts/tembo-ai/values.yaml
sjmiller609: It might be nice to add envs like this https://github.com/tembo-io/app-deploy-dev/blob/main/dev/tembo-ai.yaml

since then it's easier to just swap envs when we need. And this is subject to the list merging issue in yaml, i.e. can't merge in defaults from values file, rather the entire list needs defined if any value overwritten in the app deploy.

Also, I thought it is assuming the path in secrets store, so I feel that should be an added config instead of a default perhaps

```
            - name: DATABASE_URL
              valueFrom:
                secretKeyRef:
                  name: data-plane-inference
                  key: _data-plane_inference_db_connection_string
```

.github/workflows/helm-test.yaml
sjmiller609, line 54:
```suggestion
```

charts/tembo-ai/values.yaml
sjmiller609: That would be an issue if we use a statefulset. In the case of RWX or ROX that is a different type of storage class that typically works by like a NFS mount, in the case of AWS that's "EFS". We don't currently have that storage class in our cluster, but it would be straightforward to add.

In the case of using that kind of storage class, I think we should instead do a Deployment instead of a Statefulset, and mount all the pods to the same storage class.

charts/tembo-ai/values.yaml
ChuckHend: I see, ok that works for me. I thought this was replacing the private app-deploy process, my mistake.

PR #830, opened 2024-06-03
Add ai-proxy app to RAG stack

sjmiller609:


tembo-stacks/src/stacks/specs/rag.yaml
```diff
@@ -8,6 +8,28 @@ images:
   16: "standard-cnpg:16-389a437"
 stack_version: 0.1.0
 appServices:
+  - image: quay.io/tembo/ai-proxy:2024.06.03
```
sjmiller609, line 11: I think needs a GUC like this one

```
  - name: vectorize.embedding_service_url
    value: http://${NAMESPACE}-embeddings.${NAMESPACE}.svc.cluster.local:3000/v1/embeddings
```

The host for ai proxy is

```
${NAMESPACE}-ai-proxy:8080
```

ChuckHend (approved):

PR #835, opened 2024-06-05
Remove last_fully_reconciled_at from status

sjmiller609:


tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -877,11 +877,6 @@ async fn pods_to_fence(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<String>, A
             .as_ref()
             .and_then(|s| s.first_recoverability_time.as_ref())
             .is_none()
-        && cdb
-            .status
-            .as_ref()
-            .and_then(|s| s.last_fully_reconciled_at.as_ref())
-            .is_none()
```
sjmiller609, line 884: @nhudson WDYT about this change

nhudson: Since you're removing it from the `CoreDB` API, the `CoreDB` validation will not be able to complete.  So for each one it reads it will throw an error.

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -746,6 +746,7 @@ pub struct CoreDBStatus {
     pub runtime_config: Option<Vec<PgConfig>>,
     pub first_recoverability_time: Option<DateTime<Utc>>,
     pub pg_postmaster_start_time: Option<DateTime<Utc>>,
+    #[deprecated(note = "This field is deprecated and it is no longer used")]
```
sjmiller609, line 749: marked as deprecated instead of deleting to avoid breaking API change

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -880,7 +880,7 @@ async fn pods_to_fence(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<String>, A
         && cdb
             .status
             .as_ref()
-            .and_then(|s| s.last_fully_reconciled_at.as_ref())
+            .and_then(|s| s.pg_postmaster_start_time.as_ref())
```
sjmiller609, line 883: This value is also not set until it has come up once

nhudson (approved):
LGTM!

PR #837, opened 2024-06-06
tembo ai app

ChuckHend:


tembo-stacks/src/apps/embeddings.yaml
```diff
@@ -57,11 +57,11 @@ trunk_installs:
   - name: pg_cron
     version: 1.6.2
   - name: pgmq
-    version: 1.1.1
```
ChuckHend, line 60: embeddings app should have been previously updated, but it was not

tembo-stacks/src/apps/ai.yaml
```diff
@@ -0,0 +1,63 @@
...
+        version: 0.16.0
+postgres_config:
+  - name: cron.host
+    value: /controller/run
+  - name: vectorize.host
+    value: postgresql:///postgres?host=/controller/run
+  - name: vectorize.tembo_service_url
+    value: http://${NAMESPACE}-ai-proxy:8080/v1
+  - name: vectorize.tembo_jwt
+    value: placeholder
```
jasonmp85, line 63: What does the `value: placeholder` mean here?

jasonmp85 (approved):
One question about the placeholder values, but otherwise looks good.

tembo-stacks/src/apps/ai.yaml
ChuckHend: setting it to a "not null" for Tembo cloud users.

PR #839, opened 2024-06-07
tembo-stacks: fix parsing storages in tebibytes

vrmiguel:
* Fixes parsing `storage` decls that can now be `1Ti`, `1.5Ti` and `2Ti`
* Removes a couple of heap allocations in the process
* Removes the `regex` dependency, which could perhaps save us a bit of build time, uses a simpler naive alternative instead

### DIFF
--- a/tembo-stacks/src/stacks/config_engines.rs
+++ b/tembo-stacks/src/stacks/config_engines.rs
@@ -1,11 +1,11 @@
+use std::ops::Not;
+
 use anyhow::Result;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 use crate::stacks::types::Stack;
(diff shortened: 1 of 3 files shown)


ChuckHend (approved):

tembo-stacks/src/stacks/config_engines.rs
```diff
@@ -438,12 +441,73 @@ mod tests {
...
+            postgres_metrics: None,
+            postgres_config: None,
+            postgres_config_engine: None,
+            infrastructure: Some(Infrastructure {
+                cpu: "1".into(),
+                memory: "1Gi".into(),
+                storage: "10Gi".into(),
+            }),
+            app_services: None,
+        };
```
ChuckHend, line 489: btw - `Stack::default()` should also work here

PR #840, opened 2024-06-07
log instance_id from gateway requests

ChuckHend:
- adds a ci workflow
- logs the instance id header from all requests
- adds an integration test and some test tooling


.github/workflows/tembo_ai.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - '.github/workflows/tembo_ai.yaml'
+    - 'inference-gateway/**'
+
+jobs:
+  tests:
+    name: Run tests
+    runs-on:
+      - self-hosted
+      - dind
+      - xlarge-16x16
```
sjmiller609, line 34: is 16x16 needed? If not can pick smaller with "small" or "large-8x8". Also should include label "amd64"

sjmiller609 (approved):

.github/workflows/tembo_ai.yaml
ChuckHend: Thanks, should be able to run this on something smaller. Where can I full all the enum values for these labels?

PR #852, opened 2024-06-24
Write Traefik resources to both APIs

sjmiller609:
Part of upgrading Traefik involves switching Kubernetes API versions. For that, we write both APIs before cutting over to the new version of Traefik, then delete the old version after the rollout is done.

Other than in controller.rs, the files are either generated by kopium, or duplicates of other existing files, with different imports from the new types.


tembo-operator/src/ingress_v3.rs
```diff
@@ -0,0 +1,341 @@
+// TODO: Delete after API migration
```
sjmiller609, line 1: this file is duplicate of ingress.rs, with different types imported

nhudson (approved):

PR #853, opened 2024-06-24
operator: don't requeue for 300s if there's no Trunk project version set

vrmiguel:


sjmiller609 (approved):
I think looks good, let's wait to merge until tomorrow morning

PR #855, opened 2024-06-26
WIP Azure Backups

ianstanton:


conductor/src/main.rs
```diff
@@ -670,36 +680,63 @@ async fn init_cloud_perms(
...
+        let write_path = read_msg
+            .message
+            .backups_write_path
+            .clone()
+            .unwrap_or(format!("{}", read_msg.message.namespace));
+        destination_path = format!(
+            "https://{}.blob.core.windows.net/{}/{}",
+            azure_storage_account, backup_archive_bucket, write_path
+        );
+    }
```
sjmiller609, line 739: I think this may need a step to provision in Azure the permissions for the workload required for this instance to access the bucket where its backups go

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -89,6 +89,84 @@ pub struct ServiceAccountTemplate {
     pub metadata: Option<ObjectMeta>,
 }
 
+/// AzureCredentials is the type for the credentials to be used to upload files to Azure Blob Storage.
+#[derive(Serialize, Deserialize, Clone, Debug, Default, JsonSchema)]
+pub struct AzureCredentials {
```
sjmiller609, line 94: are these types already present in the cnpg file, or could be kopium imported from the CNPG crd

conductor/src/main.rs
ianstanton: :+1: Yep we will need that

tembo-operator/src/apis/coredb_types.rs
ianstanton: Yes they are already present. The S3Credentials  values (right below this section) are also present there, but it looks like we copied them in and renamed slightly. I followed the same pattern here. Happy to go with either

ianstanton:
Dry run
```
    backup:
      azureCredentials:
        storageAccount:
          key: ""
          name: iantembobackups
      destinationPath: https://iantembobackups.blob.core.windows.net/tembo-backups/v2/merrily-copious-herring
      encryption: AES256
      endpointURL: null
      retentionPolicy: "30"
      s3Credentials: null
      schedule: 9 9 * * *
      volumeSnapshot:
        enabled: false

```

ianstanton:
Closing for now but coming back to this soon :)

PR #856, opened 2024-06-26
inference-gateway validation

ChuckHend:
- refactor of some of the webserver startup logic so that testing can be more consistent with how webserver actually runs
- in-memory cache of the validation table. see `test_validation` for an example of the behavior
- pg_cron refresh of validation table based on a fdw

### DIFF
--- a/inference-gateway/docker/Dockerfile.postgres
+++ b/inference-gateway/docker/Dockerfile.postgres
@@ -1,0 +1,12 @@
+FROM quay.io/tembo/tembo-local
+
+RUN trunk install pg_timeseries
+RUN trunk install pg_cron
+RUN trunk install pg_partman
+RUN trunk install postgres_fdw
+
+USER root
+
+RUN echo "shared_preload_libraries = 'pg_cron'" >> $PGDATA/postgresql.conf
+
+USER postgres
--- a/inference-gateway/migrations/20240626175841_fdw.sql
+++ b/inference-gateway/migrations/20240626175841_fdw.sql
@@ -1,0 +1,29 @@
+CREATE EXTENSION IF NOT EXISTS postgres_fdw;
+CREATE EXTENSION IF NOT EXISTS pg_cron;
+
+CREATE SERVER IF NOT EXISTS foreign_validation_server
+FOREIGN DATA WRAPPER postgres_fdw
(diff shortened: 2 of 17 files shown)


inference-gateway/docker/Dockerfile.postgres
```diff
@@ -0,0 +1,12 @@
+FROM quay.io/tembo/tembo-local
+
+RUN trunk install pg_timeseries
```
sjmiller609, line 3: Perhaps version lock all the trunk installs

inference-gateway/migrations/20240626175841_fdw.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE SERVER IF NOT EXISTS foreign_validation_server
+FOREIGN DATA WRAPPER postgres_fdw
+OPTIONS (host 'localhost', port '5432', dbname 'postgres');
+
+CREATE USER MAPPING IF NOT EXISTS FOR postgres
+SERVER foreign_validation_server
+OPTIONS (user 'postgres', password 'postgres');
+
+CREATE FOREIGN TABLE IF NOT EXISTS fdw_paid_organizations (
```
sjmiller609, line 12: It seems problematic in my opinion to work this way because the data plane should not have access to the control plane db at all. For example it would cause problems with installing this into a self hosted situation.

sjmiller609 (changes requested):
I’m not sure we should go with this plan to fdw from inference gateway db to control plane

inference-gateway/docker/Dockerfile.postgres
ChuckHend: This is local dev tooling only.

inference-gateway/migrations/20240626175841_fdw.sql
ChuckHend: self hosted installations will not use organization level validation nor token-based billing models. they will set `ORG_VALIDATION_ENABLED=false`

sjmiller609: Adam mentioned on Slack we can configure this to connect to the queue DB, not the control plane DB

sjmiller609 (approved):

ChuckHend:
closing in favor of https://github.com/tembo-io/tembo/pull/858

PR #857, opened 2024-06-27
reconciler(install): find latest version of extension if not supplied

vrmiguel:


sjmiller609 (approved):
It would be ideal to include a test reproducing the regression, but this looks very safe

PR #858, opened 2024-06-27
inference-gateway validation

ChuckHend:
Sets up the logic for validating requests based on a locally populated `org_validation` table. Which is disabled by default.

- refactor of some of the webserver startup logic so that testing can be more consistent with how webserver actually runs
- in-memory cache of the validation table. see test_validation for an example of the behavior


inference-gateway/src/config.rs
```diff
@@ -20,6 +22,15 @@ impl Config {
             server_port: from_env_default("WEBSERVER_PORT", "8080")
                 .parse::<u16>()
                 .unwrap_or(8080),
+            org_validation_enabled: from_env_default("ORG_VALIDATION_ENABLED", "false")
+                .parse()
+                .expect("ORG_VALIDATION_ENABLED must be a boolean"),
```
ChuckHend, line 27: org validation disabled. And there is nothing in place to populate the local validation table.

jasonmp85 (approved):
Looks good to me/discussed in Zoom.

PR #867, opened 2024-07-10
Ensure OLAP config has enough workers for pg_cron

jasonmp85:
At the moment there's an extra bgw getting created (due to trunk modifying shared_preload_libraries), and this interaction prevents cron jobs from ever running on instances with small core counts (i.e. ones where this function returns 1).

This ensures there will be an extra worker for cron no matter what.


ChuckHend (approved):

ChuckHend (approved):
Remember to bump version in Cargo.toml

PR #877, opened 2024-07-12
Scale down poolers if hiberated

nhudson:
If a `CoreDB` has a connection pooler enabled, we should scale down the `Pooler` if `CoreDB.spec.stop` is `true`

* Add logic to scale down `Pooler` type if pause/hibernation is enabled
* Modify logic to better handle disabling of backups if pause/hibernation is enabled.

Fixes: [CLOUD-1017](https://linear.app/tembo/issue/CLOUD-1017/pooler-not-spinning-down-when-instance-is-hibernated)

### DIFF
--- a/tembo-operator/src/cloudnativepg/hibernate.rs
+++ b/tembo-operator/src/cloudnativepg/hibernate.rs
@@ -219,31 +219,38 @@ async fn update_pooler_instances(
     match pooler {
         Some(p) => {
             let current_instances = p.spec.instances.unwrap_or(1);
-            let desired_instances = if cdb.spec.stop { 0 } else { 1 };
+            let desired_instances = get_pooler_instances(cdb);
 
-            if current_instances != desired_instances {
-                let patch_pooler_spec = json!({
-                    "spec": {
-                        "instances": desired_instances,
-                    }
-                });
+            if let Some(desired) = desired_instances {
+                if current_instances != desired {
+                    let patch_pooler_spec = json!({
+                        "spec": {
+                            "instances": desired,
+                        }
+                    });
 
-                match patch_pooler_merge(cdb, ctx, patch_pooler_spec).await {
-                    Ok(_) => {
-                        info!(
-                            "Updated Pooler instances for {} from {} to {}",
-                            name, current_instances, desired_instances
-                        );
+                    match patch_pooler_merge(cdb, ctx, patch_pooler_spec).await {
+                        Ok(_) => {
+                            info!(
+                                "Updated Pooler instances for {} from {} to {}",
+                                name, current_instances, desired
+                            );
+                        }
+                        Err(e) => {
+                            error!("Failed to update Pooler instances for {}: {:?}", name, e);
+                            return Err(requeue_normal_with_jitter());
+                        }
                     }
-                    Err(e) => {
-                        error!("Failed to update Pooler instances for {}: {:?}", name, e);
-                        return Err(requeue_normal_with_jitter());
-                    }
+                } else {
+                    debug!(
+                        "Pooler instances for {} already set to {}. No update needed.",
+                        name, current_instances
+                    );
                 }
             } else {
-                debug!(
-                    "Pooler instances for {} already set to {}. No update needed.",
-                    name, current_instances
+                warn!(
+                    "Could not determine desired instances for Pooler {}. Skipping update.",
+                    name
                 );
             }
         }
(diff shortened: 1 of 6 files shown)


tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -63,7 +78,7 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
     // Conversely, setting it back to 1 if the cluster is started should reverse
     // the process.
 
-    let replicas = if cdb.spec.stop { 0 } else { 1 };
+    let replicas = get_pooler_instances(cdb);
```
bonesmoses, line 81: I don't really like this. It's making an implicit assumption that poolers are the _only_ possible Deployment. Yes, the `get_pooler_instances` function does effectively the same thing, but the name of the function is inherently misleading.

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -193,3 +208,98 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
...
+    pooler: &Option<Pooler>,
+    cdb: &CoreDB,
+    ctx: &Arc<Context>,
+) -> Result<(), Action> {
+    let name = cdb.name_any();
+
+    match pooler {
+        Some(p) => {
+            let current_instances = p.spec.instances.unwrap_or(1);
+            let desired_instances = if cdb.spec.stop { 0 } else { 1 };
```
bonesmoses, line 222: Shouldn't this be a call to `get_pooler_instances`?

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -138,25 +153,25 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
...
+    }
 
-    // Check if scheduled_backup_suspend_status=false and cdb.spec.stop=true.  If so then patch the scheduled backup suspend status to true
-    if scheduled_backup_suspend_status != cdb.spec.stop {
-        patch_scheduled_backup_merge(cdb, ctx, patch_scheduled_backup_spec.clone()).await?;
-        info!(
-            "Toggled scheduled backup suspend of {} to '{}'",
-            name, scheduled_backup_value
+    // Patch the Pooler deployment to match the hibernation state
+    if let Err(action) = update_pooler_instances(&pooler, cdb, ctx).await {
```
bonesmoses, line 166: Do we care that this is going to create a potential mismatch between `spec.connectionPooler: enabled` and the amount of requested instances?

tembo-operator/src/cloudnativepg/hibernate.rs
nhudson: I dont understand.  The value in the spec for `Pooler` is `spec.instances`, so I am getting the number of instances it should be set to.  The function call does this.  I could also just do `if cdb.spec.stop { 0 } else { 1 }`, but I figured I would create a function since it needs to be figured out more than once.

tembo-operator/src/cloudnativepg/hibernate.rs
nhudson: Thanks I missed updating that.

tembo-operator/src/cloudnativepg/hibernate.rs
bonesmoses: This section of code is for Deployments though, not pooler instances. Reusing that function here is misleading.

tembo-operator/src/cloudnativepg/hibernate.rs
nhudson: How could that happen?  If you're talking about the number of postgres instances, then this doesn't touch that.  There can only ever be 1 pooler, or at least that is all we currently support.  

`update_pooler_instances` doesn't even touch the `CoreDB`, so it never will conflict with anything.

tembo-operator/src/cloudnativepg/hibernate.rs
nhudson: Ah yeah, I understand what you're saying now.  Yes I updated the incorrect line.  This is reverted now.

tembo-operator/src/cloudnativepg/hibernate.rs
nhudson: Maybe I understand?  You're talking about the comment maybe?  If so I just updated the wording. It's patching the `Pooler` cluster resource, not the `Deployment` of the Pooler since that is handed by CNPG.

ianstanton (approved):

sjmiller609 (approved):

PR #879, opened 2024-07-15
Handle float metric values, parsing to int

sjmiller609:


nhudson (approved):

vrmiguel (approved):

conductor/src/metrics.rs
```diff
@@ -65,8 +65,9 @@ pub mod prometheus {
 
                 let timestamp = f64_val.trunc() as i64;
                 let parsed_int = str_val
```
ianstanton, line 67: Looks good but we should rename this to `parsed_float`

ianstanton (approved):
One nit

PR #881, opened 2024-07-19
Log based metrics

sjmiller609:


conductor/src/metrics_reporter.rs
```diff
@@ -80,45 +83,73 @@ pub async fn run_metrics_reporter() -> Result<()> {
...
-            .query(&[("query", query)])
-            .send()
-            .await?;
+    pub async fn query(&self, query: &str, server_type: &ServerType) -> Result<Metrics> {
+        let query_url = match server_type {
+            ServerType::Prometheus => &self.prometheus_url,
+            ServerType::Loki => &self.loki_url,
+        };
+
+        let mut request = self.client.get(query_url).query(&[("query", query)]);
```
sjmiller609, line 140: I should add a timeout, and also change the code so that any given metrics query having an issue does not cause all of the subsequent queries to fail to process

vrmiguel: Right now I think it won't stop subsequent queries, right? Will just log and move on

vrmiguel (approved):

conductor/src/metrics_reporter.rs
sjmiller609: It looks to me like a failed query will stop the loop
```
let metrics = client.query(&metric.query, &metric.server).await?;
```

PR #883, opened 2024-07-24
Collect server_name label to metrics

sjmiller609:
Collect the entire server name instead of parsing the subdomain


bonesmoses (approved):
Looks like it _vastly_ simplifies the metric query. Also, nice server names.

PR #897, opened 2024-08-05
Tembo AI billing reporting

Jayko001:


inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,146 @@
...
+    .fetch_all(dbclient)
+    .await?;
+
+    Ok(rows_to_events(rows))
+}
+
+pub fn rows_to_events(rows: Vec<UsageData>) -> Vec<Events> {
+    rows.into_iter()
+        .map(|row| Events {
+            idempotency_key: Uuid::new_v4().to_string(),
```
sjmiller609, line 56: this should be deterministic from the time range being reported on. for example if reporting for 1 hour of data, it should represent which hour (in UTC) that was being reported on.

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,146 @@
...
+    queue.init().await?;
+    queue.create(&metrics_events_queue).await?;
+
+    let mut sync_interval = interval(Duration::from_secs(3600));
+
+    loop {
+        sync_interval.tick().await;
+
+        let end_time = Utc::now();
+        let start_time = end_time - chrono::Duration::hours(1);
```
sjmiller609, line 92: Instead of doing now, minus and hour, I think we can do "for each of the previous hours in the last 24 hours not yet reported... then report the billing for that hour".

To remember which hours have already been sent, it can remember the idempotency keys in the AI database new table. The idempotency key can represent an hour of billing data. Also it should not send any hour of billing data for which the hour is not yet over.

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub message: Vec<Events>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct Events {
+    pub idempotency_key: String,
+    pub organization_id: String,
+    pub instance_id: String,
+    pub payload: Payload,
+}
```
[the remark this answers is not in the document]
sjmiller609: so I think that means the idempotency_key looks like `<instance_id>-<model>-<timestamp-to-the-hour>`

sjmiller609:
When adding the new background worker, you can add a new file called `daemon.rs`, similar to in control plane. Then add to cargo.toml to create a new entrypoint, similar to in control plane, find under `[[bin]]`

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,146 @@
...
+          organization_id,
+          instance_id,
+          duration_ms,
+          prompt_tokens,
+          completion_tokens,
+          model,
+          completed_at
+        FROM inference.requests
+        WHERE
+          completed_at >= $1 AND completed_at < $2
```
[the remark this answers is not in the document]
ChuckHend: when users check their utilization, does it only update once per hour?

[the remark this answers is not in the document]
Jayko001: as per the current model, yes.

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,158 @@
...
+    end_time: DateTime<Utc>,
+) -> Result<Vec<Events>, DatabaseError> {
+    let rows = sqlx::query_as!(
+        UsageData,
+        r#"
+        SELECT
+        organization_id,
+        instance_id,
+        model,
+        MAX(completed_at) AS completed_at,
```
vrmiguel, line 36: I could switch to COALESCE here to avoid `SUM` and `MAX` returning optional fields, not sure which option is more readable though

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    let pg_conn_url = env::var("POSTGRES_QUEUE_CONNECTION")
+        .with_context(|| "POSTGRES_QUEUE_CONNECTION must be set")?;
+    let dbclient = connect(&pg_conn_url, 5).await?;
+
+    let queue = PGMQueueExt::new(pg_conn_url, 5).await?;
+
+    // TODO: Need to set this env variable
+    let metrics_events_queue =
+        env::var("BILLING_EVENTS_QUEUE").expect("BILLING_EVENTS_QUEUE must be set");
```
sjmiller609, line 137: I think the first one is supposed to be the tembo AI DB so the env name and variable name seems off, and also I think it should not have a queue associated to it. The second one the variable name seems off. If I understand correctly, this should connect to the AI db and the billing queue only

sjmiller609 (commented):
Looking good, just one thing about the connections

sjmiller609 (approved):

inference-gateway/src/events_reporter.rs
```diff
@@ -0,0 +1,303 @@
...
+    // Run once per hour
+    const SYNC_PERIOD: Duration = Duration::from_secs(60 * 60);
+    const BILLING_QUEUE: &str = "billing_aws_data_1_use1";
+
+    let pool = db::connect(&pg_conn, 2).await?;
+
+    let queue = PGMQueueExt::new(pg_conn, 2).await?;
+
+    queue.init().await?;
+    queue.create("").await?;
```
ChuckHend, line 135: is there supposed to be a name for the queue?

PR #901, opened 2024-08-06
update vllm version in build

ChuckHend:


jasonmp85 (approved):
One character diff, how could I not approve (looks good)…

PR #902, opened 2024-08-07
Tembo Toml Struct

joshuajerin:


sjmiller609 (commented):
Is this new struct redundant of `pub struct InstanceSettings`?

joshuajerin:
Yes it is. I made a simplified struct and the types in Extension and Postgres Configurations to implement it from the Instance struct in control plane

tembo-cli/src/cmd/apply.rs
```diff
@@ -971,7 +971,7 @@ fn get_extensions_controller(
                 database: String::new(),
                 schema: None,
                 version: None,
-                enabled: extension.enabled,
+                enabled: extension.enabled.is_some(),
```
vrmiguel, line 974: I'm not quite following 🤔 

So if `extension.enabled` is `Some(false)`, it still gets set to `true`?

sjmiller609: I agree with vini, this should be unwrap or false

joshuajerin: Yes you're completely right. Changing it now

sjmiller609 (approved):

PR #908, opened 2024-08-10
Tembo Import

joshuajerin:
- Adding a Import argument that allows you to pull a cloud instance locally


tembo-cli/src/cmd/import.rs
```diff
@@ -0,0 +1,114 @@
...
+    println!("Instance imported successfully.");
+
+    Ok(())
+}
+
+// Move it to tembo client
+async fn fetch_toml(org_id: &str, instance_id: &str, config: &Configuration) -> Result<String> {
+    let client = reqwest::Client::new();
+    let url = format!(
+        "http://localhost:8080/api/v1/orgs/{}/instances/{}/toml",
```
sjmiller609, line 89: reading this from the context file

sjmiller609 (approved):
Adding a test that creates an instance, get from the endpoint, check it's same toml? Maybe include one extension

PR #911, opened 2024-08-13
Bump Tembo-CLI Version - 0.20.6

joshuajerin:


tembo-cli/src/cmd/apply.rs
```diff
@@ -944,7 +944,7 @@ fn get_extensions(
                 vec![ExtensionInstallLocation {
                     database: Some("postgres".to_string()),
                     schema: None,
-                    version: version,
+                    version: Some(version),
```
vrmiguel, line 947: `version` is already an `Option<String>`, so you can do this:

```suggestion
                    version,
```

tembo-cli/src/cmd/apply.rs
```diff
@@ -1002,7 +1002,7 @@ fn get_trunk_installs(
             if extension.trunk_project.is_some() {
                 vec_trunk_installs.push(TrunkInstall {
                     name: extension.trunk_project.unwrap(),
-                    version: version,
+                    version: Some(version),
```
vrmiguel, line 1005: This one too
```suggestion
                    version,
```

vrmiguel (approved):

PR #912, opened 2024-08-13
CI Testing for Dataplane

joshuajerin:


dataplane-webserver/tests/integration_test.rs
```diff
@@ -1,155 +1,69 @@
...
+async fn test_get_secret_names_v1() {
+    let app = create_test_app().await;
+
+    let req = test::TestRequest::get()
+        .uri("/api/v1/orgs/org_2T7FJA0DpaNBnELVLU1IS4XzZG0/instances/inst_1696253936968_TblNOY_6/secrets")
+        .to_request();
+    
+    let resp = test::call_service(&app, req).await;
+
+    assert!(resp.status().is_success(), "Response status is not successful: {:?}", resp.status());
```
vrmiguel, line 34:
```suggestion
            let status_code = resp.status();
            if status_code.is_success().not() {
                let bytes = test::read_body(resp).await;
                let msg = str::from_utf8(&bytes).unwrap();

                panic!("Request failed: {} - {msg}", status_code.as_u16())
            }
```

I'd suggest something like this so we know what the actual error is

PR #922, opened 2024-08-19
Add Analytics stack

jasonmp85:
It's OLAP (non-duckdb), with some extensions removed and pg_analytics (née pg_lakehouse) added in.


tembo-stacks/src/stacks/specs/analytics.yaml
```diff
@@ -0,0 +1,61 @@
+name: Parade Analytics
```
ChuckHend, line 1:
```suggestion
name: Analytics
```

Will be easier if it's one word no space.

jasonmp85: Thought this was a display name for some reason; fixed it.

ChuckHend (approved):

PR #924, opened 2024-08-20
add exception to egress policy for port 8080

nhudson:
There needs to be an exception to the `allow-system-egress` policy to allow for access to new Internal load balancer for the tembo-ai gateway.

We should allow access to `10.0.0.0/8` on port `8080`.  This will allow the `ai-proxy` pods access to the load balancer for the inference-gateway on all clusters.

### DIFF
--- a/tembo-operator/src/network_policies.rs
+++ b/tembo-operator/src/network_policies.rs
@@ -311,6 +311,46 @@ pub async fn reconcile_network_policies(client: Client, namespace: &str) -> Resu
 
     apply_network_policy(namespace, &np_api, allow_proxy_to_access_tembo_ai_gateway).await?;
 
+    let allow_proxy_to_access_tembo_ai_gateway_internal_lb = serde_json::json!({
+        "apiVersion": "networking.k8s.io/v1",
+        "kind": "NetworkPolicy",
+        "metadata": {
+          "name": "allow-proxy-to-access-tembo-ai-gateway-internal-lb",
+          "namespace": format!("{namespace}"),
+        },
+        "spec": {
+          "podSelector": {
+            "matchLabels": {
+              "app": format!("{}-ai-proxy", namespace)
+            }
+          },
+          "policyTypes": ["Egress"],
+          "egress": [
+            {
+              "ports": [
+                {
+                  "port": 8080,
+                  "protocol": "TCP"
+                }
+              ],
+              "to": [
+                {
+                  "ipBlock": {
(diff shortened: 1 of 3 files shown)


sjmiller609 (changes requested):
This network policy is too permissive, as it grants wide IP range access to all workloads in a customer namespace. We need to carefully sandbox the network for customer-controlled workloads. I think it would be OK to apply a policy like this on the other network policy that applies only to the ai-proxy

nhudson:
> This network policy is too permissive, as it grants wide IP range access to all workloads in a customer namespace. We need to carefully sandbox the network for customer-controlled workloads. I think it would be OK to apply a policy like this on the other network policy that applies only to the ai-proxy

Yes thanks.  I missed that specific policy.  I just updated it and it should be good now, hopefully.

tembo-operator/src/network_policies.rs
```diff
@@ -290,25 +290,23 @@ pub async fn reconcile_network_policies(client: Client, namespace: &str) -> Resu
...
                         {
-                            "namespaceSelector": {
-                                "matchLabels": {
-                                    "kubernetes.io/metadata.name": "tembo-ai"
-                                }
-                            },
-                            "podSelector": {
-                                "matchLabels": {
-                                    "app.kubernetes.io/name": "tembo-ai-gateway"
-                                }
```
sjmiller609, line 303: Does removing this block break the path to connect via cluster IP within the same cluster to the tembo-ai services?

nhudson: It does. I can add it back if needed to allow for both

sjmiller609 (approved):
I am not sure that all these statements work together in a single network policy, it might have to be separate network policies. I would try applying kind of configuration this manually in dev by disabling the reconcile on an instance, then apply this configuration, then see if it works. I think there is a pretty good chance this does not work because I recall there might be issues with specifying both ports and / or IP ranges in the same policy as namespace-selector based policies.

nhudson:
> I am not sure that all these statements work together in a single network policy, it might have to be separate network policies. I would try applying kind of configuration this manually in dev by disabling the reconcile on an instance, then apply this configuration, then see if it works. I think there is a pretty good chance this does not work because I recall there might be issues with specifying both ports and / or IP ranges in the same policy as namespace-selector based policies.

Ah yeah that is correct.  I have created a new policy with what worked when testing.

sjmiller609 (approved):

PR #928, opened 2024-08-23
update ai, mq stacks

ChuckHend:
pgvector      -> 0.7.4
pgmq           -> 1.4.2
vectorize     -> 0.18.0 (removes `/embeddings` from the `vectorize.embedding_service_url` value)
vectorscale -> 0.3.0

### DIFF
--- a/tembo-stacks/Cargo.lock
+++ b/tembo-stacks/Cargo.lock
@@ -2471,7 +2471,7 @@ dependencies = [
 
 [[package]]
 name = "tembo-stacks"
-version = "0.14.0"
+version = "0.14.1"
 dependencies = [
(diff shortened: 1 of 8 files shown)


jasonmp85 (approved):
Mostly compared whether the database version upgrades matched the trunk ones in the same file, and the parity was right there, so looks good to me.

vrmiguel (approved):

ChuckHend:
> Mostly compared whether the database version upgrades matched the trunk ones in the same file, and the parity was right there, so looks good to me.

I'm going to make a todo to add a check that confirms all of the requested extension versions actually exist in trunk (for the trunk_installs).

PR #938, opened 2024-09-04
Add support for Google Cloud

nhudson:
Adding GCP configuration support in the operator

* Updating tests to use kind v1.29 (1.28 in AWS, 1.29 in GCP)
* Update tests to use the same version of Traefik we do in AWS/GCP
* Add `Backup` configuration options to support GKE Workload Identity Auth or passing in GCP credentials.
* Appended new `NetworkPolicy` to allow for DNS lookups in GCP.  Keeps compatibility with AWS.

fixes: [CLOUD-1217](https://linear.app/tembo/issue/CLOUD-1217/[operator]-add-support-for-deploying-instances-to-gcp)


nhudson:
Conductor tests will fail, since we are testing against a released version of the operator and not the version in this branch.

tembo-operator/src/network_policies.rs
```diff
@@ -39,42 +39,68 @@ pub async fn reconcile_network_policies(client: Client, namespace: &str) -> Resu
         "apiVersion": "networking.k8s.io/v1",
         "kind": "NetworkPolicy",
         "metadata": {
-          "name": "allow-egress-to-kube-dns",
-          "namespace": format!("{namespace}"),
+            "name": "allow-egress-to-gke-dns",
```
ianstanton, line 42: Is there a reason for the name change? It looks like we're adding one egress entry for GKE but the existing entry for kube-dns is still there.

ianstanton (approved):
Awesome! Just one question about the netpol

tembo-operator/src/network_policies.rs
nhudson: The GKE clusters are built with node-local-dns enabled.  This runs as a `DaemonSet` on each node and intercepts DNS traffic to `kube-dns`.  So the policy was modified to allow traffic to either `node-local-dns` or allow traffic to `kube-dns` both running in the `kube-system` namespace.

This allows the policy to work for both EKS and GKE (with or without node-local-dns enabled).

ianstanton: That makes sense. I think it we should give this netpol a more generic name as it will be present on GKE, EKS and eventually AKS clusters. What do you think?

nhudson: I will add, if we do need to disable `node-local-dns` for any reason then the policy will still work as expected, and we can just remove it later.  There are pro's and con's for running it and not running it.  `kube-dns` has some limitations (20 open connections per pod).  GKE does auto scale `kube-dns` but that can lead to other issues at scale.  `node-local-dns` takes that load off of `kube-dns` under heavy load and will cache DNS queries.  This leads to faster DNS responses, but sometimes can cause other issues on larger clusters.  

After some reading if this happens, we should migrate to using CloudDNS with GKE which I haven't seen before, but it's the recommended path https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-dns.  This comes with some added cost to running the GKE cluster as well.

nhudson: Oh I see what you're talking about @ianstanton.  I will make that change to make it more generic.

sjmiller609 (approved):
Nick and I discussed to be careful about compatibility with existing resources, since the CRD is being updated.

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -222,14 +268,21 @@ pub struct Backup {
...
     #[serde(default, rename = "endpointURL")]
     pub endpoint_url: Option<String>,
 
     /// The S3 credentials to use for backups (if not using IAM Role)
     #[serde(default = "defaults::default_s3_credentials", rename = "s3Credentials")]
     pub s3_credentials: Option<S3Credentials>,
 
+    /// The Google Cloud Storage credentials to use for backups
+    #[serde(
+        default = "defaults::default_google_credentials",
```
sjmiller609, line 281: Maybe we remove this default, since that may include the default google creds in the case of AWS backups

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -530,8 +589,7 @@ pub struct CoreDBSpec {
     /// backups and WAL archive uploads to an S3 compatible object store.
     ///
     /// **Default**: disabled
-    #[serde(default = "defaults::default_backup")]
-    pub backup: Backup,
+    pub backup: Option<Backup>,
```
sjmiller609, line 592: maybe we roll this back to simplify this PR to not worry about incompatible CRDs

sjmiller609 (commented):
In a peer programming session, we are determining how to roll this out

sjmiller609 (approved):

PR #943, opened 2024-09-05
update vllm

ChuckHend:
https://github.com/vllm-project/vllm/releases/tag/v0.6.0

### DIFF
--- a/.github/workflows/tembo_ai.yaml
+++ b/.github/workflows/tembo_ai.yaml
@@ -52,15 +52,13 @@ jobs:
       - name: install dependencies
         run: cargo install sqlx-cli --version 0.8.2
       - name: Start Postgres
-        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/tembo_ai.yaml
```diff
@@ -50,17 +50,15 @@ jobs:
         with:
           toolchain: stable
       - name: install dependencies
-        run: cargo install sqlx-cli --version 0.7.4
```
ChuckHend, line 53: updating sqlx-cli version, and running the tests using the sqlx cache.  Running the tests using sqlx offline, and running migrations executed from webserver code.  Then putting `make check` at the end after the tests, so that we don't have to explicitly run the migrations.

ianstanton (approved):

PR #946, opened 2024-09-09
Create and manage Workload Identity Policies in GCP

nhudson:


conductor/src/main.rs
```diff
@@ -40,36 +41,54 @@ const REQUEUE_VT_SEC_SHORT: i32 = 5;
 // that we would want to try again after awhile.
 const REQUEUE_VT_SEC_LONG: i32 = 300;
 
+fn get_env_var(key: &str, default: Option<&str>) -> Result<String, ConductorError> {
+    Ok(env::var(key).unwrap_or_else(|_| default.unwrap_or("").to_owned()))
+}
```
sjmiller609, line 46: I think we should crash when required env vars are missing

conductor/src/main.rs
```diff
@@ -40,36 +41,54 @@ const REQUEUE_VT_SEC_SHORT: i32 = 5;
...
+    })?;
+    let gcp_project_id = get_env_var("GCP_PROJECT_ID", None)?;
+    let gcp_project_number = get_env_var("GCP_PROJECT_NUMBER", None)?;
+
+    // Error and exit if both IS_CLOUD_FORMATION and IS_GCP are set to true
+    if is_cloud_formation && is_gcp {
+        return Err(ConductorError::ConnectionPoolError(
+            "Cannot have both IS_CLOUD_FORMATION and IS_GCP set to true".into(),
+        ));
+    }
```
sjmiller609, line 84: in cases of invalid configuration, just crash the program with the error, so it's impossible to start it

conductor/src/main.rs
```diff
@@ -40,36 +41,54 @@ const REQUEUE_VT_SEC_SHORT: i32 = 5;
...
+
+    // Error and exit if IS_GCP is true and GCP_PROJECT_ID or GCP_PROJECT_NUMBER are not set
+    if is_gcp && (gcp_project_id.is_empty() || gcp_project_number.is_empty()) {
+        return Err(ConductorError::ConnectionPoolError(
+            "GCP_PROJECT_ID and GCP_PROJECT_NUMBER must be set if IS_GCP is true".into(),
+        ));
+    }
```
sjmiller609, line 91: same here

conductor/src/main.rs
```diff
@@ -420,6 +450,21 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
                 info!("{}: Deleting cloudformation stack", read_msg.msg_id);
                 delete_cloudformation(aws_region.clone(), &namespace).await?;
```
sjmiller609, line 451: Nick mentioned need to handle not doing this line unless AWS backups were enabled. Maybe by skipping it unless it's AWS

sjmiller609 (commented):
A few clean-up items but I think it looks good. Will leave unapproved until un-draft

ianstanton (approved):
This is looking ready for some QA in dev :)

PR #948, opened 2024-09-10
Add `extraEnv` to Conductor Chart Values

ianstanton:
- Add `extraEnv` value to conductor chart
- Include in all deployment templates


shahadarsh (approved):
LGTM :thumbsup:

PR #958, opened 2024-09-13
update argo deployment for usc1 gke cluster

nhudson:
Update Argo deployments to deploy to usc1 GKE cluster for prod only


shahadarsh (approved):
Spoke through this with Nick. LGTM.

PR #961, opened 2024-09-17
deletes cf stack after instance is stopped

shahadarsh:
This is done so that the `iam-role` created using cloud formation gets deleted if the instance is stopped(paused) to avoid reaching the `iam-role` per account limit. It will get recreated when the instance is started back up.

### DIFF
--- a/conductor/src/main.rs
+++ b/conductor/src/main.rs
@@ -430,10 +430,8 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
                 let spec_js = serde_json::to_string(&current_spec.spec).unwrap();
                 debug!("dbname: {}, current_spec: {:?}", &namespace, spec_js);
 
-                if read_msg.message.event_type == Event::Stop {
-                    if is_cloud_formation {
-                        let status = current_spec.status.clone().unwrap();
-
+                if is_cloud_formation && read_msg.message.event_type == Event::Stop {
+                    if let Some(status) = current_spec.clone().status {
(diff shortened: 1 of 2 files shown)


conductor/src/main.rs
```diff
@@ -430,6 +430,29 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
                 let spec_js = serde_json::to_string(&current_spec.spec).unwrap();
                 debug!("dbname: {}, current_spec: {:?}", &namespace, spec_js);
 
+                if read_msg.message.event_type == Event::Stop {
```
shahadarsh, line 433: Didn't move this into its own function because there are too many params that gets passed to the function and it becomes messy. We should at some point refactor the entire function and break it down into multiple smaller functions.

sjmiller609 (changes requested):

conductor/src/main.rs
```diff
@@ -430,6 +430,27 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
                 let spec_js = serde_json::to_string(&current_spec.spec).unwrap();
                 debug!("dbname: {}, current_spec: {:?}", &namespace, spec_js);
 
+                if is_cloud_formation && read_msg.message.event_type == Event::Stop {
+                    if let Some(status) = current_spec.clone().status {
```
shahadarsh, line 434: Is there a better way to do this?

sjmiller609 (approved):

PR #963, opened 2024-09-18
Fix for Dedicated Networking

joshuajerin:
- Allow IP Allow List to work with Dedicated Networking
- Creating 2 more hostname if Dedicated Networking is enabled (1 if DN enabled, 1 if Standby is enabled) to not disrupt the original one


tembo-operator/src/dedicated_networking.rs
```diff
@@ -356,10 +356,16 @@ async fn reconcile_dedicated_networking_service(
         service_name, namespace, service_type, lb_scheme
     );
 
+    let external_dns_hostname = if is_standby {
+        format!("dedicated-ro.{}.{}", namespace, basedomain)
+    } else {
+        format!("dedicated.{}.{}", namespace, basedomain)
+    };
```
vrmiguel, line 363: Alternatively you could do something like this:

```rust
    let external_dns_hostname = format!(
        "dedicated{prefix}.{namespace}.{basedomain}",
        prefix = if is_standby { "-ro" } else { "" }
    );
```

Whichever you prefer

joshuajerin: Oh nice! Thank you

vrmiguel (approved):

tembo-operator/tests/integration_tests.rs
```diff
@@ -2168,7 +2168,7 @@ mod test {
             .as_ref()
             .expect("Annotations should be present");
         let basedomain = std::env::var("DATA_PLANE_BASEDOMAIN").unwrap();
-        let expected_hostname = format!("{}.{}", namespace, basedomain);
+        let expected_hostname = format!("dedicated.{}.{}", namespace, basedomain);
```
ianstanton, line 2171: Let's add one for when `is_standby == true`, where `expected_hostname` would equal `format!("dedicated-ro.{}.{}", namespace, basedomain)`

ianstanton (changes requested):
Just a few changes requested! Can you also add some details to the description?

ianstanton:
I think we have some flaky tests in the `tembo-operator` functional tests. I have seen this one failing in my branches as well: `test::functional_test_restart_and_update_replicas`

ianstanton (approved):

PR #968, opened 2024-09-23
Add support for Azure Backups

ianstanton:
Add support for Azure backup and restore in `tembo-operator` and `conductor`.

Conductor changes include steps for creating the Azure 'IAM' resources necessary for instances to perform backup / restore with Azure Blob Storage.

When an instance using Azure is created:
- Creates User Assigned Managed Identity
- Creates Role Assignment
- Creates Federated Credentials
- Passes relevant info as part of the CoreDB spec

When an instance using Azure is deleted:
- Deletes the User Assigned Managed Identity

_Note: This only supports the backup archive bucket at the moment. We'll add support for the storage archive bucket as a follow-up_


nhudson (approved):

ianstanton:
We need to add this spec value to the Cluster resource for Azure deployments:
```
  inheritedMetadata:
    labels:
      azure.workload.identity/use: "true"
```
I'll follow up with this shortly.

PR #974, opened 2024-09-25
Added background worker settings for pg_cron

shhnwz:


jasonmp85:
I know you may be worried about the effects of removing the `host` setting, but if removing it has no effect, I _really_ think it would be best. Dead config will just cause confusion to everyone who reads these files in the future.

ChuckHend:
we may need to update how we set `cron.max_running_jobs` before this merges

PR #977, opened 2024-09-25
tembo-cli: fix clippy + update tembo-stacks and controller crates

ChuckHend:
- Updates `controller` and `tembo-stacks` crates to latest version
- fixes bunch of clippy warnings
- adds implementation for ai_proxy app type

This does not fix the failing test though.


joshuajerin (approved):

shahadarsh (approved):
LGTM. Test is failing but somehow its using the same instance name between tests.

PR #979, opened 2024-09-25
Handle GCP restore URI destination path correctly

nhudson:
Now that we have GCP we need to be able to set and apply the correct `destination_path` instead of just using `s3://`.

fixes:[CLOUD-1325](https://linear.app/tembo/issue/CLOUD-1325/restores-are-failing-because-restore-pod-is-pulling-image-from-ecr)


conductor/src/lib.rs
```diff
@@ -35,18 +35,30 @@ pub async fn generate_spec(
...
+
     // Add the bucket name into the backups_path if it's not already there
     if let Some(restore) = &mut spec.restore {
         if let Some(backups_path) = &mut restore.backups_path {
-            if !backups_path.starts_with(&format!("s3://{}", backups_bucket)) {
-                let path_suffix = backups_path.trim_start_matches("s3://");
-                *backups_path = format!("s3://{}/{}", backups_bucket, path_suffix);
+            if !backups_path.starts_with(&format!("{}{}", prefix, backups_bucket)) {
+                let path_suffix = backups_path.trim_start_matches(trim_prefix);
+                *backups_path = format!("{}{}/{}", prefix, backups_bucket, path_suffix);
```
sjmiller609, line 57: this code reads like it's trimming from the start of the string the same thing that it adds right back to it. It also seems confusing. I think the logic is clearer like this:

- Remove any prefix matching any of the known prefixes from the backups path
- Format the string with the prefix, given we know which cloud we are running on

conductor/src/lib.rs
```diff
@@ -35,18 +35,30 @@ pub async fn generate_spec(
     namespace: &str,
     backups_bucket: &str,
     spec: &CoreDBSpec,
-) -> Value {
+) -> Result<Value, ConductorError> {
     let mut spec = spec.clone();
+    let cloud_provider = data_plane_id.split('_').next().unwrap_or("");
```
sjmiller609, line 40: I think parsing the cloud provider from the data plane ID is not the best way to know which cloud provider is being used

sjmiller609 (changes requested):

conductor/src/lib.rs
nhudson: I am open to suggestions to how to work around it.  Currently the only data plane related info we get in Conductor is set as the `data_plane_id`.  If that isn't the best way maybe we could add a cloud provider string in the message from the control plane?  

We could do a look up the dataplane from the database, but we still couldn't use the `data_plane_id` to look that up since it's a string from the fields in the table.  Maybe just passing the provider id from the control-plane in the message is the right idea?

ianstanton: Thinking about this from a self hosted perspective as well. In those cases we would need to add something that would allow for choosing only one cloud provider based on the environment it's been installed in.

nhudson: In that case we already set the environment variable of `is_cloudformation` or `is_gcp` in Conductor.  We will need something like that for Azure as well.  So I think we can just assume the cloud provider from those settings.

If both of you are fine with that I can get that added.

ianstanton: I like that idea!

conductor/src/lib.rs
nhudson: Thanks.  I believe I have changed it to your specifications.

conductor/src/lib.rs
nhudson: The last commit addresses this and set the cloud provider based off the env var that is set.

conductor/src/lib.rs
```diff
@@ -35,18 +40,38 @@ pub async fn generate_spec(
...
-                *backups_path = format!("s3://{}/{}", backups_bucket, path_suffix);
+
+    match cloud_provider {
+        CloudProvider::AWS | CloudProvider::GCP => {
+            let prefix = cloud_provider.prefix();
+
+            // Format the backups_path with the correct prefix
+            if let Some(restore) = &mut spec.restore {
+                if let Some(backups_path) = &mut restore.backups_path {
+                    let clean_path = remove_known_prefixes(backups_path);
```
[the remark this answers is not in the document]
nhudson: It shouldn't but it's basically doing what it was already doing I think.

We did strip the `s3://` prefix and prepended the `s3://<default bucket name>`, which is what is happening here as well.  Just with the specific cloud provider.

This support both `s3://<bucket path>` and `s3://<bucket-name>/<bucket-path>` and sets them correctly. 

The logic isn't changing and we already had tests, testing both scenarios.  It's just changed to do the exact same thing but set the correct provider URI.  Hopefully this makes sense.

ianstanton (approved):

sjmiller609 (approved):

PR #980, opened 2024-09-26
`conductor`: Add support for Azure Backup and Restore

ianstanton:
This PR adds steps for creating the Azure 'IAM' resources necessary for instances to perform backup / restore with Azure Blob Storage.

When an instance using Azure is created:
- Creates User Assigned Managed Identity
- Creates Role Assignment
- Creates Federated Credentials
- Passes relevant info as part of the `CoreDB` spec

When an instance using Azure is deleted:
- Deletes the User Assigned Managed Identity

_Note: This only supports the backup archive bucket at the moment. We'll add support for the storage archive bucket as a follow-up_


conductor/src/lib.rs
```diff
@@ -586,6 +592,74 @@ pub async fn delete_gcp_storage_workload_identity_binding(
...
+    let uami_name = namespace;
+    let uami = create_uami(
+        azure_resource_group,
+        azure_subscription_id,
+        azure_region,
+        uami_name,
+        credentials.clone(),
+    )
+    .await?;
+    info!("Created UAMI: {:?}", uami);
```
ianstanton, line 616: These are here for testing. I'll update with more appropriate logs.

nhudson (changes requested):
Might be good to add a check like these to make sure you can't set aws or gcp also at the same time.  Other than that LGTM!  Great work!

```
    // Error and exit if both IS_CLOUD_FORMATION and IS_GCP are set to true
    if is_cloud_formation && is_gcp {
        panic!("Cannot have both IS_CLOUD_FORMATION and IS_GCP set to true");
    }

    // Error and exit if IS_GCP is true and GCP_PROJECT_ID or GCP_PROJECT_NUMBER are not set
    if is_gcp && (gcp_project_id.is_empty() || gcp_project_number.is_empty()) {
        panic!("GCP_PROJECT_ID and GCP_PROJECT_NUMBER must be set if IS_GCP is true");
    }
```

ianstanton:
> Might be good to add a check like these to make sure you can't set aws or gcp also at the same time. Other than that LGTM! Great work!
> 
> ```
>     // Error and exit if both IS_CLOUD_FORMATION and IS_GCP are set to true
>     if is_cloud_formation && is_gcp {
>         panic!("Cannot have both IS_CLOUD_FORMATION and IS_GCP set to true");
>     }
> 
>     // Error and exit if IS_GCP is true and GCP_PROJECT_ID or GCP_PROJECT_NUMBER are not set
>     if is_gcp && (gcp_project_id.is_empty() || gcp_project_number.is_empty()) {
>         panic!("GCP_PROJECT_ID and GCP_PROJECT_NUMBER must be set if IS_GCP is true");
>     }
> ```

Thanks, good suggestion! Addressed here https://github.com/tembo-io/tembo/pull/980/commits/a4b9cd5ef84393c7a6ba4a43f764afad8a0e1d08

nhudson (approved):

PR #983, opened 2024-09-26
Fix disabling of VolumeSnapshots when instances are disabled

nhudson:
Currently the `VolumeSnapshots` are not disabled when the instance is paused/hibernated.  This leads to backed up `Backups` being created that will cause an issue with backups and snapshots when the instance is un-paused.

Fixes: [CLOUD-1340](https://linear.app/tembo/issue/CLOUD-1340/volumesnapshots-not-disabled-when-instance-is-pausedhibernated)

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -47,10 +47,10 @@ jobs:
           set -xe
           sudo apt-get update
           sudo apt-get install -y pkg-config libssl-dev git
-      - uses: azure/setup-helm@v3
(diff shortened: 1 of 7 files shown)


nhudson:
I wanted to write a test, but there is no way to test for the `ScheduledBackup` for snapshots since we can't enable snapshots on a Kind cluster.

nhudson:
Tested this in dev aws use1 and it's working as expected.  Now just :crossed_fingers: that the flaky tests pass.

ianstanton (approved):

PR #993, opened 2024-10-02
ai: multi inference services

ChuckHend:


ChuckHend:
replacing with https://github.com/tembo-io/tembo/pull/994

PR #999, opened 2024-10-03
cli(minimal-cloud): don't run `tembo apply` twice

vrmiguel:


tembo-cli/src/cli/context.rs
```diff
@@ -102,11 +103,11 @@ impl Profile {
     }
 }
 
-impl ToString for Target {
-    fn to_string(&self) -> String {
+impl Display for Target {
```
vrmiguel, line 106: Clippy lint

joshuajerin (approved):
THANK YOU

PR #1002, opened 2024-10-08
 ai-billing-reporter: update billing watermark

vrmiguel:


inference-gateway/migrations/20241008000306_add-id-to-last-reported-at.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE billing.reporter_watermark ADD COLUMN id INT DEFAULT 1 UNIQUE;
```
ChuckHend, line 1:
```suggestion
ALTER TABLE billing.reporter_watermark ADD COLUMN id INT DEFAULT 1 UNIQUE;
ALTER TABLE billing.reporter_watermark ALTER COLUMN last_reported_at TIMESTAMP WITH TIME ZONE DEFAULT '1970-01-01' NOT NULL;

```

PR #1003, opened 2024-10-08
Conductor feature for Dedicated Networking

joshuajerin:


conductor/src/main.rs
```diff
@@ -85,6 +85,10 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
         .unwrap_or_else(|_| "".to_owned())
         .parse()
         .expect("error parsing GCP_PROJECT_NUMBER");
+    let is_loadbalancer_public: bool = env::var("IS_LOADBALANCER_PUBLIC")
+        .unwrap_or_else(|_| "false".to_owned())
```
nhudson, line 89: Only a question.  We talked about making the default `true` here.  Having it `false` is fine just wanted to make sure that is what you intended.

shahadarsh: Agreed. Let's make it true by default.

nhudson (approved):

conductor/src/main.rs
shahadarsh: I would also change operator to have public true by default

tembo-operator/src/dedicated_networking.rs
```diff
@@ -399,12 +399,6 @@ async fn reconcile_dedicated_networking_service(
         "cnpg.io/cluster".to_string(),
         serde_json::Value::String(cdb_name.to_string()),
     );
-    if is_public {
```
shahadarsh, line 402: Is this not needed anymore?

joshuajerin: I don't think so. We have "service.beta.kubernetes.io/aws-load-balancer-internal: "false" for that

tembo-operator/src/dedicated_networking.rs
```diff
@@ -427,8 +421,23 @@ async fn reconcile_dedicated_networking_service(
     service_spec.insert("type".to_string(), json!(service_type));
     let ip_allow_list = cdb.spec.ip_allow_list.clone().unwrap_or_else(|| vec![]);
 
-    if service_type == "LoadBalancer" && !ip_allow_list.is_empty() {
-        service_spec.insert("loadBalancerSourceRanges".to_string(), json!(ip_allow_list));
+    // Allow ip_allow_list to allow all entries are in CIDR notation
```
shahadarsh, line 424: Were you able to test this locally? If not please do that in dev.

tembo-operator/src/dedicated_networking.rs
shahadarsh: Ok Thanks.

tembo-operator/src/dedicated_networking.rs
joshuajerin: Tested it locally. Will test it out again in dev once its merged

PR #1009, opened 2024-10-16
allow model rewrites

ChuckHend:


.github/workflows/cargo-test.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
         with:
           contains_the_file: Cargo.toml
           changed_relative_to_ref: origin/${{ github.base_ref || 'not-a-branch' }}
-          ignore_dirs: ".coredb examples tembo-cli/temboclient tembo-cli/tembodataclient"
+          ignore_dirs: ".coredb examples tembo-cli/temboclient tembo-cli/tembodataclient inference-gateway"
```
ChuckHend, line 28: this project has its own CLI file, `tembo-ai.yml` that run tests and clippy etc with its own configurations.

PR #1015, opened 2024-10-17
Add wget to kind installation in actions

ianstanton:
Fixes failures like: https://github.com/tembo-io/tembo/actions/runs/11387443430/job/31682060161?pr=968


nhudson (approved):

ianstanton:
Operator workflow failure seems unrelated to dependencies. Will look into it if we still see if after merging this one.

PR #1018, opened 2024-10-18
init sql runner app

DarrenBaldwin07:
- This adds a new SQL runner app (opt-in, so not turned on for any stacks by default yet)
- App source is here: https://github.com/tembo-io/sql-runner


tembo-stacks/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "tembo-stacks"
 description = "Tembo Stacks for Postgres"
-version = "0.17.2"
+version = "0.17.3"
```
ChuckHend, line 4: Since we are adding a completely new feature but still backwards compatible.

```suggestion
version = "0.18.0"
```

ChuckHend (approved):

PR #1023, opened 2024-10-21
conductor: add heartbeat monitor for background workers

vrmiguel:


conductor/src/heartbeat_monitor.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[derive(Clone)]
+pub struct HeartbeatMonitor {
+    shared_heartbeat: Arc<AtomicU64>,
+    update_interval: Duration,
+}
+
+#[derive(Clone)]
+pub struct HeartbeatUpdater {
+    shared_heartbeat: Arc<AtomicU64>,
```
vrmiguel, line 22: Went with a lock-free approach rather than something like `Arc<RwLock<Instant>>`,  as the issue with workers getting stuck might stem from lock contention, so adding more lock contention probably wouldn't help

conductor/src/heartbeat_monitor.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+}
+
+impl HeartbeatUpdater {
+    pub fn update_heartbeat(&self) {
+        self.shared_heartbeat
+            .store(current_timestamp(), Ordering::Relaxed);
+    }
+}
```
vrmiguel, line 63: I think it's better to keep this as something manually updated rather than being updated by yet another background thread

conductor/src/heartbeat_monitor.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Checks if the heartbeat is still active
+    ///
+    /// # Returns true if the heartbeat has been updated within the expected time frame, false if the heartbeat has not been updated within twice the expected timeout duration
+    pub fn is_heartbeat_active(&self) -> bool {
+        let last_update = self.shared_heartbeat.load(Ordering::Relaxed);
+        let current_time = current_timestamp();
+
+        if current_time >= last_update {
+            let elapsed = Duration::from_secs(current_time - last_update);
+            elapsed < self.update_interval * 2
```
ChuckHend, line 50: what is the `* 2` part for?

vrmiguel: To check if there's been an update within twice the expected timeout duration

vrmiguel: Maybe we could replace `self.update_interval` by `timeout_interval` and then just use. `elapsed < self.timeout_interval`, what do you think?

PR #1025, opened 2024-10-22
Revert `pg_search` to 0.10.3

ianstanton:
Revert `pg_search` extension to `0.10.3` in ParadeDB stack.


nhudson:
I guess this is not needed now?

PR #1028, opened 2024-10-25
Add `ScheduledDeletion` State

joshuajerin:


joshuajerin:
From testing in dev, we should be able to restore an instance without the coredbspec and namespace

conductor/src/main.rs
```diff
@@ -549,61 +549,69 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
                 info!("{}: Deleting namespace {}", read_msg.msg_id, &namespace);
                 delete_namespace(client.clone(), &namespace).await?;
 
-                if is_cloud_formation {
-                    info!("{}: Deleting cloudformation stack", read_msg.msg_id);
-                    delete_cloudformation(aws_region.clone(), &namespace).await?;
-                }
+                if read_msg.message.event_type == Event::Delete {
```
nhudson, line 552: IMHO this if statement isn't needed.  If the instance is paused, we have already removed the IAM/Workload Idenitity/UAMI from the cloud provider anyway.  It's safe to run for both events (`Delete` and `ScheduledDeletion`).

nhudson (approved):

conductor/src/types.rs
```diff
@@ -33,6 +33,8 @@ pub enum Event {
     Started,
     Restore,
     Restored,
+    ScheduleDeletion,
+    ScheduleDeletionComplete,
```
ChuckHend, line 37: Make sure these align exactly with what the control-plane is expecting when serialized.

joshuajerin:
We decided not to add any new states so I'm closing this ticket

PR #1031, opened 2024-11-01
Make sure we cleanup any VolumeSnapshots after a set retention peroid

nhudson:
This PR adds some code to cleanup `VolumeSnapshots` and `Backup` objects from the cluster after the defined `volume_snapshot_retention_peroid` config value.

* Adds the `volume_snapshot_rentention_period` value with a default of 40 days.  This is in line with our current retention for object store backups in AWS and GCP.
* Checks for all `Backup` method types of `volumeSnapshot` older than `volume_snapshot_rentention_period`.  Then looks up the associated `VolumeSnapshot`
* Deletes the `Backup` then will delete the `VolumeSnapshot` if successful.

This is to fix an issue we are seeing with the snapshot-controller handling too many snapshots on the cluster and in our cloud services.  CNPG currently doesn't support deleting snapshots after some retention period.

fixes: [PRO-2209](https://linear.app/tembo/issue/PRO-2209/snapshot-controller-timing-out-when-querying-for-all-objects-at)

### DIFF
--- a/tembo-operator/src/config.rs
+++ b/tembo-operator/src/config.rs
@@ -4,6 +4,7 @@ use std::env;
 pub struct Config {
     pub enable_backup: bool,
     pub enable_volume_snapshot: bool,
+    pub volume_snapshot_rentention_period: u64,
     pub reconcile_timestamp_ttl: u64,
     pub reconcile_ttl: u64,
 }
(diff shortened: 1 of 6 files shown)


tembo-operator/src/config.rs
```diff
@@ -15,6 +16,12 @@ impl Default for Config {
             enable_volume_snapshot: from_env_default("ENABLE_VOLUME_SNAPSHOT", "false")
                 .parse()
                 .unwrap(),
+            volume_snapshot_rentention_period: from_env_default(
+                "VOLUME_SNAPSHOT_RETENTION_PERIOD",
```
ChuckHend, line 20: Can we put units on the env var name? Like `VOLUME_SNAPSHOT_RETENTION_PERIOD_HRS` or whatever the units are.

nhudson: It's in days.  Yeah good idea.  I will make that change.

ChuckHend (approved):

shahadarsh (approved):

PR #1033, opened 2024-11-04
fix lookup of backup objects in the cluster

nhudson:
With #1031 changes there is an error when looking up `Backups` with using a `ListParam`.  This changes how we filter for `volumeSnapshot` method backups.

```rust
2024-11-04T18:05:45.955686Z ERROR reconciling object:reconcile:reconcile:cleanup_old_volume_snapshots: controller::cloudnativepg::retention::snapshots:63: Failed to list backups in namespace barely-permissive-deer: ApiError: field label not supported: spec.method: BadRequest (ErrorResponse { status: "Failure", message: "field label not supported: spec.method", reason: "BadRequest", code: 400 }) object.ref=CoreDB.v1alpha1.coredb.io/barely-permissive-deer.barely-permissive-deer object.reason=object updated trace_id=00000000000000000000000000000000 instance=barely-permissive-deer namespace=barely-permissive-deer retention_days=40
```

### DIFF
--- a/tembo-operator/src/cloudnativepg/retention/snapshots.rs
+++ b/tembo-operator/src/cloudnativepg/retention/snapshots.rs
@@ -52,25 +52,32 @@ pub async fn cleanup_old_volume_snapshots(
         "Starting cleanup of old volume snapshots for instance: {}", name
     );
 
-    // List only volume snapshot backups
-    let lp = ListParams::default().fields(&format!("spec.method={}", BackupMethod::VolumeSnapshot));
-
-    // List all backups in the namespace
-    let backups = backups_api
-        .list(&lp)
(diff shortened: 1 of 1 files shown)


shahadarsh (approved):

jasonmp85 (approved):

nhudson:
> What's the bug preventing the API filter from working?
> 
> Code change looks good to me for in-memory workaround, I'm just curious why it broke…

Great question.  The `spec.method` field is not a field that `ListParams` supports to look up objects with.  I knew this, but did it anyway and never thought about it until I saw the error.

The `ListParams` method only supports looking up valid `labels` from the metadata, not actual fields inside the API object.

PR #1038, opened 2024-11-11
Stop CNPG reconciliation for hibernated instances

vrmiguel:


tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -142,18 +145,26 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
         }
     }
 
-    // Build the hibernation patch we want to apply to disable the CNPG cluster.
+    // Stop CNPG reconciles for hibernated instances
```
ChuckHend, line 148: let's add a note here about why it is important that we dont set turn off reconciliation until hibernation is complete.

nhudson (approved):

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -333,3 +344,22 @@ async fn update_scheduled_backups(
 
     Ok(())
 }
+
+fn is_cluster_hibernated(cluster: &Cluster) -> bool {
+    fn get_hibernation_condition(cluster: &Cluster) -> Option<&ClusterStatusConditions> {
```
ChuckHend, line 349: can we get a test in place for this fn?

PR #1042, opened 2024-11-14
`reconcile_cnpg`: don't exec into Pod if it's not yet ready

vrmiguel:


tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1055,6 +1059,20 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
...
+        Ok(true) => {
+            // Pod is ready, continuing
+        }
+        Ok(false) => {
+            info!("In reconcile_cnpg: Pod is not ready, requeuing 10 seconds");
+            return Err(Action::requeue(Duration::from_secs(10)));
+        }
+        Err(err) => {
+            error!("Failed to determine if pod is ready: {err}");
+            return Err(Action::requeue(Duration::from_secs(30)));
```
nhudson, line 1073: Lets change this to 300 seconds.  If there is an error it should requeue after a longer period of time.

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1055,6 +1059,20 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
...
         }
     }
+    // Don't exec into the pod if it's not Ready yet
+    match is_pod_ready(&primary_pod_cnpg) {
+        Ok(true) => {
+            // Pod is ready, continuing
+        }
+        Ok(false) => {
+            info!("In reconcile_cnpg: Pod is not ready, requeuing 10 seconds");
+            return Err(Action::requeue(Duration::from_secs(10)));
```
nhudson, line 1069: What do you think about changing this to 30 seconds?  I just don't want to hit the K8s API every 10 seconds waiting on a pod that will never come up.  (eg: Full PVC).

30 seconds isn't great either but I feel like we should quickly try again for pods that eventually will come up.

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1047,6 +1047,10 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
             None
         }
     };
+    // If we can't find the existing primary pod, returns a requeue
+    let primary_pod_cnpg = cdb
+        .primary_pod_cnpg_ready_or_not(ctx.client.clone())
```
nhudson, line 1052: Ok I was worried that this would be an issue, but since we are calling `primary_pod_cnpg_ready_org_not` which calls this function https://github.com/tembo-io/tembo/blob/cac0c8960e4312abfe430a7f793987bf9b4dd832/tembo-operator/src/controller.rs#L525 we should be fine.  Since it will return the pod even if it's not ready.

This part is GTG

nhudson (changes requested):

nhudson (approved):

PR #1043, opened 2024-11-15
update GCP storage bucket conditions to allow for restores to work

nhudson:
Update the GCP Storage Bucket condition to allow restores to work in GCP

fixes: [PRO-2249](https://linear.app/tembo/issue/PRO-2249/[gcp]-update-iam-condition-to-allow-for-restores-to-work)

PR #1044, opened 2024-11-15
Revert "`reconcile_cnpg`: don't exec into Pod if it's not yet ready"

nhudson:
Reverts tembo-io/tembo#1042

Reverts change made in #1042 .  This will cause new `CoreDB`'s to never setup and place the `Cluster` in the namespace.


nhudson:
Adding fix for this revert in #1043, so closing this.

PR #1052, opened 2024-11-21
update conductor helm chart, fix resource group names

nhudson:
* Update conductor helm chart
  * Support external secrets from Azure Key Vault
  * Support adding `podLabels` so that the pod will use Azure Workload Identity
  
* Update conductor code to rename the resource groups where the infrastructure lives.

### DIFF
--- a/charts/conductor/values.yaml
+++ b/charts/conductor/values.yaml
@@ -12,6 +12,9 @@ externalSecrets:
   parameterStore:
     name: "secret-store-parameter-store"
     kind: ClusterSecretStore
+  # remoteRef:
+  #   key: ~
+  #   secretKey: ~
   secretName: ~
   secretRegex: ~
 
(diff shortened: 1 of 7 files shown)


charts/conductor/values.yaml
```diff
@@ -12,6 +12,9 @@ externalSecrets:
   parameterStore:
     name: "secret-store-parameter-store"
     kind: ClusterSecretStore
+  # remoteRef:
+  #   key: ~
+  #   secretKey: ~
```
ChuckHend, line 17: not needed anymore or was this for debug?

nhudson: Its commented out for reference.  We accept the values if they exist and don't if they don't.

nhudson: I will add a comment explaining.

ChuckHend (approved):

charts/conductor/values.yaml
ChuckHend: cool just checking!

PR #1058, opened 2024-11-25
Remove `IngressRouteTCP` for Stopped Instances

joshuajerin:
https://linear.app/tembo/issue/TEM-2241/%5Btembo-operator%5D-when-paused-we-should-remove-the-ingressroutetcp-for

### DIFF
--- a/tembo-operator/src/cloudnativepg/hibernate.rs
+++ b/tembo-operator/src/cloudnativepg/hibernate.rs
@@ -153,7 +153,7 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) -> 
         delete_ingress_route_tcp(ingress_route_tcp_api.clone(), &namespace, &prefix_read_only).await
     {
         warn!(
-            "Error deleting ingress route for {}: {}",
+            "Error deleting postgres ingress route for {}: {}",
             cdb.name_any(),
             err
         );
@@ -169,7 +169,7 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) -> 
     .await
     {
         warn!(
-            "Error deleting extra ingress route for {}: {}",
+            "Error deleting postgres ingress route for {}: {}",
             cdb.name_any(),
             err
         );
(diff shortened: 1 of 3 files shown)


tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -145,6 +146,23 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
         }
     }
 
+    // Remove IngressRouteTCP route for stopped instances
+    let ingress_route_tcp_name = format!("{}-ro-0", cdb.name_any().as_str());
+    if let Err(err) = delete_ingress_route_tcp(
+        Api::namespaced(client, &namespace),
```
nhudson, line 152: I don't think this will work.  It will need to be typed, so maybe adding something like this will work better

```
let ingress_route_tcp_api: Api<IngressRouteTCP> = Api::namespaced(ctx.client.clone(), namespace);
```

Then just passing in `ingress_route_tcp_api` to the `delete_ingress_route_function`

nhudson (changes requested):

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -145,6 +146,21 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
         }
     }
 
+    // Remove IngressRouteTCP route for stopped instances
+    let ingress_route_tcp_name = format!("{}-ro-0", cdb.name_any().as_str());
```
nhudson, line 150: Where are you getting this name from?  The names would need to be looked up somehow.

For example the `IngressRouteTCP` name on most of them end with `rw-0`, but we will also need to handle any custom domain configurations as well.

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -145,6 +146,60 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
...
+    {
+        warn!(
+            "Error deleting postgres ingress route for {}: {}",
+            cdb.name_any(),
+            err
+        );
+        return Err(Action::requeue(Duration::from_secs(300)));
+    }
+
+    let prefix_extra = format!("extra-{}-rw", cdb.name_any().as_str());
```
joshuajerin, line 191: Grabbed this from here 
https://github.com/tembo-io/tembo/blob/5b154ffbfc6f0a5d64e5f6a7949fede7c29f7806/tembo-operator/src/ingress.rs#L95

nhudson, line 191: We may need to check if the extra domain is added or not.

We should look and see if it's set then delete if it is

```
 let mut extra_domain_names = cdb.spec.extra_domains_rw.clone().unwrap_or_default();
```

tembo-operator/src/cloudnativepg/hibernate.rs
```diff
@@ -145,6 +146,60 @@ pub async fn reconcile_cluster_hibernation(cdb: &CoreDB, ctx: &Arc<Context>) ->
...
+    {
+        warn!(
+            "Error deleting postgres ingress route for {}: {}",
+            cdb.name_any(),
+            err
+        );
+        return Err(Action::requeue(Duration::from_secs(300)));
+    }
+
+    let prefix_pooler = format!("{}-poller-0", cdb.name_any().as_str());
```
nhudson, line 179: This name is misspelled.  It should be `pooler` not `poller`.

nhudson (changes requested):
I believe it might also be best to add some tests for this in the intergration test.  You should be able to add some asserts in the `functional_test_hibernate` test to check if the `IngressRouteTCP` was removed or not.

PR #1082, opened 2024-12-17
Only delete ingress objects if stopped

ianstanton:
Wrap ingress deletion logic in condition check for `cdb.spec.stop`. Without this, ingress objects were deleted and re-created for _all_ instances.

### DIFF
--- a/tembo-operator/tests/integration_tests.rs
+++ b/tembo-operator/tests/integration_tests.rs
@@ -4369,19 +4369,19 @@ CREATE EVENT TRIGGER pgrst_watch
         assert!(service_items.is_empty());
         // should be no Services
 
-        // ingress must be gone
+        // There should be 1 IngressRoute left (metrics)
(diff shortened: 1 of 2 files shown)


nhudson (approved):

ChuckHend (approved):

tembo-operator/tests/integration_tests.rs
```diff
@@ -4369,19 +4369,19 @@ CREATE EVENT TRIGGER pgrst_watch
         assert!(service_items.is_empty());
         // should be no Services
 
-        // ingress must be gone
+        // There should be 1 IngressRoute left (metrics)
```
ianstanton, line 4372: These assertions were checking for 0 `IngressRoute`s and 0 `IngressRouteTCP`s. They were being deleted and re-created before introducing this fix. We expect 1 of each to be present.

PR #1096, opened 2024-12-20
Update data_warehouse stack extensions

theory:
Necessary to support Postgres 17, though it cannot be used until we can update pg_tier.


theory:
Closing at we no longer provision new clusters with this stack.

PR #1098, opened 2024-12-20
Update machine_learning stack extensions

theory:
Necessary to support Postgres 17, though it cannot be used until postgresml is updated.

### DIFF
--- a/tembo-stacks/src/stacks/specs/machine_learning.yaml
+++ b/tembo-stacks/src/stacks/specs/machine_learning.yaml
@@ -26,39 +26,39 @@ postgres_config:
     value: postgresql:///postgres?host=/controller/run
 trunk_installs:
   - name: pgvector
-    version: 0.7.4
+    version: 0.8.0
   - name: postgresml
-    version: 2.7.1
+    version: 2.9.3
   - name: pg_cron
-    version: 1.6.2
+    version: 1.6.3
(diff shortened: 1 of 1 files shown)


nhudson (approved):

ChuckHend (changes requested):
I'm almost certain pgml > 2.7.1 has new dependencies that are not in our ml image.

theory:
> I'm almost certain pgml > 2.7.1 has new dependencies that are not in our ml image.

It doesn't look like you changed anything in tembo-io/trunk@acc2c75, @ChuckHend.

ChuckHend:
> It doesn't look like you changed anything

I am referring to image that we use to run pgml on tembo-cloud. this PR will not run there.

theory:
> I am referring to image that we use to run pgml on tembo-cloud. this PR will not run there.

Ah, I assumed that since you didn't need to update dependencies to build it, then none would need to be updated in the image. I'll make a PR for the image.

ChuckHend:
I think it would be better to build pgml into the image rather than installing it from trunk. You'll be able to build the image and test it in dev before PRing here for the new image. Then this PR can be closed.

theory:
How does /tembo-io/tembo-images#77 look?

theory:
Closing at we're deprecating this stack.

PR #1099, opened 2024-12-20
Update pg_partman in the message_queue stack

theory:


ChuckHend (approved):
I assume no breaking changes right? its only here because pgmq needs it, i havent checked 5.2's behavior.

theory:
> I assume no breaking changes right? its only here because pgmq needs it, i havent checked 5.2's behavior.

Via Slack, @keithf4 tells me only bug fixes and new features.

PR #1103, opened 2024-12-20
Update timeseries stack extensions

theory:
Necessary to support Postgres 17, though it cannot be used until pg_tier and columnar are updated. Therefore remove the Postgres 17 config, which has yet to work.

Increment tembo-stacks to v0.23.9.

Completes tem-2693.

### DIFF
--- a/tembo-stacks/Cargo.lock
+++ b/tembo-stacks/Cargo.lock
@@ -2472,7 +2472,7 @@ dependencies = [
 
 [[package]]
 name = "tembo-stacks"
-version = "0.23.8"
+version = "0.23.9"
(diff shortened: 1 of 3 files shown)


ChuckHend:
pg_timeseries also depends on hydra's columnar, which does not currently support pg17, but pg17 is not in the 0.1.7 release

theory:
> pg_timeseries also depends on hydra's columnar, which does not currently support pg17, but pg17 is not in the 0.1.7 release

Thanks, updated the commit message in 89a88fc8.

nhudson (approved):

PR #1134, opened 2025-01-23
Update all the apps

theory:
Update all the app stack base images and extensions. Contributes to tem-2797.

### DIFF
--- a/tembo-stacks/src/apps/http.yaml
+++ b/tembo-stacks/src/apps/http.yaml
@@ -1,6 +1,6 @@
 name: !http
 appServices:
-  - image: postgrest/postgrest:v10.0.0
+  - image: postgrest/postgrest:v12.2.5
     name: http
     resources:
       requests:
@@ -49,10 +49,10 @@ appServices:
         value: postgres
 trunk_installs:
   - name: pg_graphql
-    version: 1.4.4
+    version: 1.5.9
 extensions:
   - name: pg_graphql
     locations:
       - database: postgres
         enabled: true
-        version: 1.4.4
+        version: 1.5.9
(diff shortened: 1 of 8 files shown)


tembo-stacks/src/apps/http.yaml
```diff
@@ -1,6 +1,6 @@
 name: !http
 appServices:
-  - image: postgrest/postgrest:v10.0.0
+  - image: postgrest/postgrest:v12.2.5
```
ChuckHend, line 3: any breaking changes from v10 -> v12? The things on my radar would be these env vars: `PGRST_DB_URI,  PGRST_DB_SCHEMA, PGRST_DB_ANON_ROLE`

tembo-stacks/src/apps/sql-runner.yaml
```diff
@@ -1,6 +1,8 @@
 name: !sqlrunner
 appServices:
-  - image: quay.io/tembo/sql-runner:c62f892
+  # Use latest commit SHA in the main branch of
+  # https://github.com/tembo-io/sql-runner.
+  - image: quay.io/tembo/sql-runner:f522d49
```
ChuckHend, line 5: @DarrenBaldwin07 - is this the image tag we want to be using?

ChuckHend (approved):

tembo-stacks/src/apps/sql-runner.yaml
DarrenBaldwin07: yes ✅

tembo-stacks/src/apps/http.yaml
theory: Dunno, will try stuff in dev once merged.

PR #1142, opened 2025-02-03
Tembo client regen

DarrenBaldwin07:
This PR:
- regenerates control-plane api client from updated openapi spec so that structs are refreshed (we generated it ~11 months ago so most types were outdated)
- moves api client to the root of repo since it is a separate crate


vrmiguel (approved):
Great work!!

PR #1187, opened 2025-03-13
Update tembo-operator for new base image

theory:
The new tembo/postgres image stores files in new locations on the persistent volume, so teach `CoreDBSpec` to recognize the new image and add the methods `pg_major`, `share_dir`, `module_dir`, and `lib_dir` to return the proper locations. Update `check_for_so_files()` to use these methods rather than a hard-coded directory, and also to include the `.so` file name in the `ls` command to reduce I/O.

Remove the old `postgres_major_version_from_cdb` function, as the `pg_major` method takes its place (and returns a default rather than an error when it can't parse the image name for some reason).

Leave various defaults and tests using the existing images, as more updates will be needed before we can run the new image, but sync them up to use the latest image SHA1, the same on found in stack configurations.

Also update `tembo-operator/yaml/sample-standard.yaml` to require the latest `pg_stat_statements`.

Also: the `RolePassword.password` field is never read, so change is name to `_password` to eliminate warnings.

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -27,7 +27,7 @@ on:
 
 jobs:
   functional_test:
-    name: Run functional testing
+    name: 🧪 Operator on ☸️ ${{ matrix.kube_version }}
     runs-on:
       - operator-runner-8x32
     strategy:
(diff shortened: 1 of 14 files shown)


nhudson (approved):

theory:
Was merged in [2befbb0](https://github.com/tembo-io/tembo/commit/2befbb089f7287cdeb5c0407f70875e0dad8365b).

PR #1192, opened 2025-03-19
Fix `check_for_so_files()`

theory:
Broken in 2befbb0, where I failed to realize that the command would error out if the file did not exist. So teach the `ls` command to ignore errors by sending STDERR to `/dev/null` and to execute `true` if `ls` returns an error.

Also from 2befbb0, comment out the `--pkglibdir` option in the call to `trunk`. It doesn't work for images that don't have a version of Trunk with that option. We can turn it back on once all images have upgraded.


nhudson (approved):

ChuckHend (approved):
🚀

PR #1214, opened 2025-04-18
fix lint errors, remove deployments to staging

nhudson:
* Remove all deployments to the staging environment
* Remove `tembo-ai` deployments as we no longer use this
* Fix linting errors that were causing build errors for `tembo-cli` and `cp-client`
* Remove deprecated secrets API path as it was causing build errors

---

- Linear: [TEM-3623](https://linear.app/tembo/issue/TEM-3623/update-tembo-repo-to-not-deploy-to-plat-staging)


nhudson:
Looks like I will need another PR to fix the `tembo-cli` build after we merge this so that it can pick up the fix from `cp-client`

theory (approved):
By secrets!

nhudson:
> Bye secrets!

We still get secrets, it's just removing the old deprecated API that isn't used anymore.

PR #1220, opened 2025-04-29
Fixes to the backup download API

nhudson:
There are quite a few fixes in here

* Remove `actix-rt` crate in favor of using `actix_web::rt::spawn` instead
* The `aws-sdk-s3` client requires to explicitly install a process-level CryptoProvider with `rustls`
* Use server-side-encryption with `put_object()` method
* Update `Dockerfile` to install `ca-certificates` and to update to `debian:12.10-slim` and `rust:1.86-bookworm`
* Rework how we install/execute the `temback` binary
* Ensure the backup is present in the backup bucket

### DIFF
--- a/dataplane-webserver/src/backups/mod.rs
+++ b/dataplane-webserver/src/backups/mod.rs
@@ -65,10 +65,17 @@ pub async fn perform_backup_task(
             .await?;
 
     // Define the metadata key
-    let metadata_key = format!("{}/temback/{}/status.json", bucket_path, job_id);
+    let metadata_key = format!("{bucket_path}/status.json");
 
     // Update the status based on the backup result
-    update_backup_status(client, &bucket_name, &metadata_key, &backup_result).await?;
+    update_backup_status(
+        client,
+        &bucket_name,
+        &metadata_key,
+        &backup_result,
+        &namespace,
+    )
+    .await?;
 
     match backup_result {
         BackupResult::Success => {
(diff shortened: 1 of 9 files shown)


nhudson:
Running the following works:

```
curl -X POST "https://api.data-1.use1.cdb-dev.com/api/v1/orgs/{org_id}/instances/{instance_id}/backup" -H 'accept: */*' -H 'Authorization: Bearer {$TEMBO_JWT}'  -H 'Content-Type: application/json'

{"job_id":"aac9faa9-e073-4139-b8ba-f2923294dc82","status":"processing"}
```

Then to check if it's completed

```
curl -X GET "https://api.data-1.use1.cdb-dev.com/api/v1/orgs/{org_id}/instances/{instance_id}/backup/aac9faa9-e073-4139-b8ba-f2923294dc82" -H 'accept: */*' -H 'Authorization: Bearer {$TEMBO_JWT}'  -H 'Content-Type: application/json'

{"status":"completed","job_id":"aac9faa9-e073-4139-b8ba-f2923294dc82","download_url":"https://cdb-plat-use1-dev-instance-backups.s3.us-east-1.amazonaws.com/v2/...","expires_at":"2025-04-30T18:50:42.497752408+00:00"}
```

dataplane-webserver/src/backups/mod.rs
```diff
@@ -280,108 +416,96 @@ async fn run_backup_on_instance_pod(
...
         "--bucket",
         &bucket_name,
         "--dir",
         &bucket_path,
+        "--pass",
+        &password,
+        "--host",
+        "localhost",
+        "--user",
+        "postgres",
```
theory, line 448: You can omit these now if it they're not actually needed to connect. Presumably if it's running inside the pod where localhost connections are trusted, you don't need any of this stuff.

Also, do we want to use `--text`?

theory (approved):

dataplane-webserver/src/backups/mod.rs
nhudson: Ok just confirmed that removing the host, user and pass work.  I don't know why we would use `--text` for this situation

vrmiguel (approved):

PR #1223, opened 2025-04-30
always install latest version of temback

nhudson:
Just in case there is ever a situation where an older version of `temback` is already installed, we should always check and update it to the correct version used in the `dataplane-webserver`

### DIFF
--- a/dataplane-webserver/src/backups/mod.rs
+++ b/dataplane-webserver/src/backups/mod.rs
@@ -16,6 +16,7 @@ use kube::{
 use tokio::io::AsyncReadExt;
 
 const TEMBACK_INSTALL_DIR: &str = "/var/lib/postgresql/data";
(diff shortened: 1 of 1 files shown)


dataplane-webserver/src/backups/mod.rs
```diff
@@ -309,36 +338,35 @@ async fn install_temback(
             stderr.read_to_string(&mut _err).await.ok();
         }
 
-        // Verify the binary exists after installation
-        let check_cmd = vec!["ls", temback_path.as_str()];
+        // Verify the binary exists and check its version
+        let verify_cmd = vec!["/var/lib/postgresql/data/temback", "--version"];
```
theory, line 342: Dupe!

theory (approved):

dataplane-webserver/src/backups/mod.rs
nhudson: Thanks.  Fixed!

